PR #649, opened 2014-11-22
Incomplete limit

dhalperi:
Implement and test a limit operator, which emits only the specified number of tuples.

Note that since we can't close upstream operators, this isn't actually a working implementation of a limit operator.

But, this is a start, and we cannot do the right thing with the current network implementation.


dhalperi:
@BrandonHaynes and here is the Myria limit operator

PR #675, opened 2015-01-01
Use the hash function for x64

jingjingwang:
Since we are mostly using 64-bit machines and according to the [doc](https://code.google.com/p/smhasher/wiki/MurmurHash3) this should give us better performance. I did a rough profiling and also have seen the improvement a little bit.

### DIFF
--- a/src/edu/washington/escience/myria/util/HashUtils.java
+++ b/src/edu/washington/escience/myria/util/HashUtils.java
@@ -23,10 +23,10 @@ public final class HashUtils {
 
   /** The hash functions. */
   private static final HashFunction[] HASH_FUNCTIONS = {
-      Hashing.murmur3_32(SEEDS[0]), Hashing.murmur3_32(SEEDS[1]), Hashing.murmur3_32(SEEDS[2]),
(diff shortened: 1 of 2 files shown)


dhalperi:
Hmm, I'm a bit confused here. There are two different things going on

1. the internal state of the hash algorithm -- 32 bits (currently used) vs 128 bits (suggested)
2. the processor width for which the algorithm is optimized. It looks like the 32-bit hasher is optimized for 32-bit CPUs and the 128-bit hasher is optimized for 64-bit CPUs. There is no 64-bit-CPU version of the 32-bit algorithm. http://smhasher.googlecode.com/svn/trunk/MurmurHash3.cpp

They do not produce equivalent results (as seen; you have changed the tests).

I guess they ought to work roughly equivalently well as hash algorithms. So the question would be, does the 32-bit internal state on 32-bit CPU perform better or worse than the 128-bit internal state on 64-bit CPUs? Might be worth making some micro benchmarks to test. E.g., one might have different caching behavior.

jingjingwang:
They did a speed test [here](https://code.google.com/p/smhasher/wiki/MurmurHash3), in the section called "Bulk speed test".

dhalperi:
great!

Issue #665, opened 2014-12-17
issues with deployment scripts


dhalperi:
- [x] `launch_cluster` if the master does not start successfully, script still tries to start the workers. This shouldn't happen.
- [x] `update_jar_only` should also update the `sqlite4java` distribution. (Maybe we should rename the script). The point is that "update_jar_only", in spirit, should be "everything that setup_cluster does, but without redoing anyone's catalog." So copying over sqlite4java is important when we upgrade the sqlite library.
- [x] `launch_cluster`: figure out how to print fewer warnings when the master is starting up.

dhalperi:
Bumped by @mbalazin:

> When I run the initial deployment, I get an exception (see below). This exception is very scary looking and makes it seem like everything failed but apparently it did not fail..

```java
WARN  2015-02-19 17:40:40,973 [main] DeploymentUtils - expected exception occurred
java.net.ConnectException: Connection refused
    at java.net.PlainSocketImpl.socketConnect(Native Method)
    at java.net.AbstractPlainSocketImpl.doConnect(AbstractPlainSocketImpl.java:339)
```

> should we change this error message to make it more obvious that this is a warning and not an error? Otherwise, people will try to debug a non-existing error. What do you think? I think that simply not printing the stack but only printing a "WARN: Trying to connect. Failed. Will retry." (or whatever the error actually means) would make it more clear that it's normal.

jingjingwang:
I think they have all been resolved so I'm closing this issue.

PR #676, opened 2015-01-01
Also sync sqlite4java

jingjingwang:
I also renamed the script name as suggested in #665 but open to other good names.


dhalperi:
Thanks!

PR #677, opened 2015-01-01
Jwang add join pull order encoding

jingjingwang:
By default it's ALTER if not specified.
ALTER: pull both children alternatively.
LEFT: for each fetchNextReady(), start from the left child, and stick to pulling the left child until no data, then switch the the right child.
RIGHT: same as LEFT.

### DIFF
--- a/src/edu/washington/escience/myria/api/MyriaExceptionMapper.java
+++ b/src/edu/washington/escience/myria/api/MyriaExceptionMapper.java
@@ -82,8 +82,7 @@ public final class MyriaExceptionMapper implements ExceptionMapper<Throwable> {
       return ((MyriaApiException) cause).getResponse();
     }
     ResponseBuilder ret = Response.status(status);
-    if (status.equals(Status.INTERNAL_SERVER_ERROR) || status.equals(Status.BAD_REQUEST) || cause.getMessage() == null
-        || cause.getMessage().length() == 0) {
+    if (status.equals(Status.INTERNAL_SERVER_ERROR) || cause.getMessage() == null || cause.getMessage().length() == 0) {
       ret.entity(ErrorUtils.getStackTrace(cause));
     } else {
       ret.entity(cause.getMessage());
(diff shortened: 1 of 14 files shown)


jingjingwang:
So the case that I met was a invalid hard-coded JSON plan. There were some wrong opIds mapping from producers to consumers, and the exceptions were eventually passed into MyriaExceptionMapper.toResponse (https://github.com/uwescience/myria/blob/master/src/edu/washington/escience/myria/api/MyriaExceptionMapper.java#L35) as a Status.BAD_REQUEST, but ex.getMessage() didn't contain much helpful information.

jingjingwang:
One example is https://github.com/uwescience/myria/blob/master/src/edu/washington/escience/myria/api/encoding/QueryConstruct.java#L242,  this gave me:
java.lang.IndexOutOfBoundsException: Index: 0, Size: 0
when the json was wrong, but I couldn't see what went wrong exactly.

dhalperi:
Makes perfect sense! As I suspected, we need a better error message there. In fact, we should be doing more checks... right?

here's a (typed in GitHub) proposal... what are your thoughts?

```java
List<ExchangePairID> outputs = producerOutputChannels.get(idbInput.opId);
Verify.verifyNotNull(outputs, "no entry for IDBInput in producerOutputChannels %s", idbInput.opId);
Preconditions.checkArgument(outputs.size() == 1, "Expected exactly 1 channel into IDBInput %s, found: %s", idbInput.opId, outputs);
// now get the first element of outputs[].
```

jingjingwang:
Yeah, I agree that adding more checks is another way to solve the problem. There are other places to add checks too, I'll add a few more and revert  95dd132.

dhalperi:
@jingjingwang any response to comments? if not, I may, later, clean up as I see fit and then merge

THanks!

jingjingwang:
@dhalperi  PS: I think I've done the fixes.

PR #679, opened 2015-01-05
RootOperator: return the coalesced batch

dhalperi:
If coalescing is enabled, the tuple batch returned by RootOperator may
have been incorrect because we returned the original, not coalesced,
batch.

Doubt this had much practical impact, but it could affect profiling results.


dhalperi:
Bug confirmed: https://demo.myria.cs.washington.edu/profile?queryId=60234# and click fragment 0. Then zoom to 8.367s -- 8.390s.

The Store purports to return 2247 tuples. However, while the first batch on the left is 2247 tuples, several more batches are coalesced before insertion -- the total is closer to 10K I think.

screen shot 2015-01-05 at 1 30 19 pm

PR #683, opened 2015-01-21
profiling of multi-subquery queries

dhalperi:
FYI @domoritz , I need this pretty bad for some of the work I'm doing and I'm thinking about how to get there.

A straw proposal that involves no frontend work is to simply add logging of events and physical plans for all subqueries and then add a subquery parameter to the frontend. I'm going to start there.

- [x] update the profiling and sent log tables

- [x] update the profiling queries and API calls

- [x] let the backend profile individual subqueries. (Rather than the workaround here: https://github.com/uwescience/myria/blob/b95cfcc05ff180bf133dea1c4b1113af6cdc3a54/src/edu/washington/escience/myria/parallel/Query.java#L182)

- [x] log the physical plans for every subquery. (Rather than the workaround here: https://github.com/uwescience/myria/blob/b95cfcc05ff180bf133dea1c4b1113af6cdc3a54/src/edu/washington/escience/myria/parallel/QueryManager.java#L491)

- [x] modify myria-web; create an issue there and link it here. See uwescience/myria-web#246

### DIFF
--- a/src/edu/washington/escience/myria/profiling/ProfilingLogger.java
+++ b/src/edu/washington/escience/myria/profiling/ProfilingLogger.java
@@ -17,6 +17,7 @@ import edu.washington.escience.myria.accessmethod.ConnectionInfo;
 import edu.washington.escience.myria.accessmethod.JdbcAccessMethod;
 import edu.washington.escience.myria.operator.Operator;
 import edu.washington.escience.myria.parallel.ResourceStats;
+import edu.washington.escience.myria.parallel.SubQueryId;
 import edu.washington.escience.myria.parallel.WorkerSubQuery;
 import edu.washington.escience.myria.storage.TupleBatch;
 import edu.washington.escience.myria.storage.TupleBatchBuffer;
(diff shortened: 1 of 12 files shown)


src/edu/washington/escience/myria/profiling/ProfilingLogger.java
```diff
@@ -181,11 +186,13 @@ public synchronized void recordEvent(final Operator operator, final long numTupl
    */
   public synchronized void recordSent(final Operator operator, final int numTuples, final int destWorkerId)
       throws DbException {
-    sent.putLong(0, operator.getQueryId());
-    sent.putInt(1, operator.getFragmentId());
-    sent.putLong(2, getTime(operator));
-    sent.putLong(3, numTuples);
-    sent.putInt(4, destWorkerId);
+    SubQueryId sq = operator.getSubQueryId();
+    events.putLong(0, sq.getQueryId());
```
domoritz: this should be `sent`, right?

dhalperi: good catch. copy and paste.

dhalperi:
@domoritz any initial thoughts on this + uwescience/myria#246. I think it's essentially G2G -- any other changes would be cosmetic.

domoritz:
LGTM, this as well as the changes to the front end in https://github.com/uwescience/myria-web/pull/246

dhalperi:
```sql
ALTER TABLE "public:logs:Profiling" ADD COLUMN "subQueryId" bigint NOT NULL default 0;
ALTER TABLE "public:logs:Profiling" ALTER "subQueryId" DROP DEFAULT;
ALTER TABLE "public:logs:Sending" ADD COLUMN "subQueryId" bigint NOT NULL default 0;
ALTER TABLE "public:logs:Sending" ALTER "subQueryId" DROP DEFAULT;
```

also drop the indexes:

```sql
DROP INDEX "public:logs__myria_indexes:Profiling_0_1_3_4";
DROP INDEX "public:logs__myria_indexes:Profiling_0_1_2_3_4";
DROP INDEX "public:logs__myria_indexes:Sending_0_1";
```

dhalperi:
In master.catalog:

```sql
CREATE TABLE query_plans (
    query_id INTEGER NOT NULL,
    subquery_id INTEGER NOT NULL,
    plan TEXT NOT NULL,
    PRIMARY KEY (query_id, subquery_id));
```

PR #686, opened 2015-01-28
Query completion speedup

dhalperi:
@domoritz identified a fairly slow transition between subqueries during yesterday's meeting. I took a guess that the issue is related to updating the catalog information about those subqueries on disk.

Here's some work that attempts to improve this behavior. Only loosely tested so far.


dhalperi:
Now the time between query finishing and query completion is on the order of 180ms, down from 400–500. My belief is this is now just the time to update the SQLite catalog.

Also note that only subqueries that write back to the Catalog are affected, and this is typically only 1 subquery per query.

PR #694, opened 2015-02-11
Redo query construction and worker mapping

### LINKED ISSUE #693: bug in query construct with multiple undefined fragments
Comes up in this query:

```myrial
T = scan(Twitter5M);
Joined = [from T A, T B, T C
          where A.$1 = B.$0 and B.$1 = C.$0 and C.$1 = A.$0
          emit A.$0 as x, B.$0 as y, C.$0 as z];
C = countall(Joined);
store(C, Triangles);
```

The last two fragments are:

1. Store(SplitConsumer)
2. SplitProducer(countall combiner(CollectConsumer)).

we have two constraints here:

a. Fragment 2 above is run on a single worker (CollectConsumer).
b. Fragment 1 above is run on the same workers as fragment 2.

However, in the current query construct code, we are not honoring the edge constraint and thus Fragment 1 is set to be all workers. Then, the query plan never terminates.

dhalperi:
I broke down the process of assigning workers to fragments
into 6 steps, with rules of precedence:

1. Obey user-overrides of fragment workers.
2. Fragments that scan tables must use the workers that contain the data.
3. Edge constraints between fragments. E.g., a `{@link LocalMultiwayProducerEncoding}` must use the same set of workers as its consumer.
4. Singleton constraints: Fragments with a `{@link CollectConsumerEncoding}` or a `{@link SingletonEncoding}` must run on a single worker. If none is still set, choose an arbitrary worker.
5. Repropagate edge constraints.
5. Unspecified: Any fragments that still have unspecified worker sets will use all workers in the cluster. (No need to repropagate edge constraints).

I then refactored the existing 2 functions that did all of this, incorrectly, into
six and heavily commented each one. There is now a lot more sanity-checking of plans
as well, and all the existing tests pass!

- [x] Fix #693.
- [x] add more tests.


dhalperi:
@jingjingwang I think you would be the best one to review this PR, if you have time. The main question marks for me have to do with my sanity checks of edges, e.g., that only MultiwayProducer and EOSController can have multiple affiliated consumers.

dhalperi:
A future PR should remove some of the redundancy in the code... I recompute a lot of things in those 6 functions, but it was the best way IMO to make it very clear what was going on.

dhalperi:
@jingjingwang bump ;)

jingjingwang:
I can imagine to have more tests in JsonBadPlanTests, but this one is good enough as a start  :)

dhalperi:
Looks good, except for this LMP/LMC issue: https://github.com/uwescience/myria/pull/694#commitcomment-10034455.

Do you have any specific tests in mind you'd like to add now? Obviously I was unable to add the tests for the EOS stuff I didn't understand :)

PR #700, opened 2015-02-19
update examples for MyriaX doc

jingjingwang:


jingjingwang:
http://myria.cs.washington.edu/docs/myriaX.html

jingjingwang:
@dhalperi  I think you're the best person to be assigned since the issue was created by you :)

src/edu/washington/escience/myria/util/DeploymentUtils.java
```diff
@@ -292,7 +293,10 @@ public static void ensureMasterStart(final String hostname, final int restPort,
           break;
         }
         // expected for the first few trials
-        LOGGER.warn("expected exception occurred", e);
```
dhalperi: Should save the exception here and, if the master never comes up, print the last one we received. Sometimes the errors are useful!

dhalperi: @jingjingwang bump -- this is why this PR is not yet merged.

jingjingwang:
@dhalperi  done :)

PR #701, opened 2015-02-24
Added configuration option to specify a database port

lbrendanl:
With this commit, myria will no longer have its postgres port hardcoded
to 5401.  Rather, it will use the default port of each database,
unless otherwise specified by the user in the configuration file.

The production cluster config file will need to add a database port setting set to 5401, since our postgres instance listens on this port, and not the default.

### DIFF
--- a/myriadeploy/myriadeploy.py
+++ b/myriadeploy/myriadeploy.py
@@ -35,6 +35,13 @@ def read_config_file(filename='deployment.cfg'):
     except ConfigParser.NoOptionError:
         ret['ssl'] = False
 
+    # Check for custom postgres port
+    try:
+        ret['database_port'] = config.get('deployment', 'database_port')
+        ret['custom_port'] = True
+    except ConfigParser.NoOptionError:
+        ret['custom_port'] = False
+
     # Helper function
     def split_hostportpathdbname_key_append_id(hostport):
         "Splits id = host,port,path,db_name into a tuple (host, port, path, db_name, id)."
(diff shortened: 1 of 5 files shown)


lbrendanl:
@dhalperi @domoritz  let me know what you think of this.

myriadeploy/myriadeploy.py
```diff
@@ -35,6 +35,13 @@ def read_config_file(filename='deployment.cfg'):
     except ConfigParser.NoOptionError:
         ret['ssl'] = False
 
+    # Check for custom postgres port
+    try:
+        ret['database_port'] = config.get('deployment', 'database_port')
+        ret['custom_port'] = True
```
domoritz: what is this used for?

domoritz:
LGTM. Just make sure the tests are not failing.

myriadeploy/myriadeploy.py
dhalperi: yeah, looks like it's not used :p

src/edu/washington/escience/myria/accessmethod/ConnectionInfo.java
```diff
@@ -102,7 +103,13 @@ public static String toJson(final String dbms, final String hostName, final Stri
         // Now it is hardcoded to use a specific connection info, which allows only one
         // myria instance per machine in the cluster
         host = hostName;
-        port = MyriaConstants.STORAGE_MONETDB_PORT;
+            
+        if (databasePort == null) {
+            port = MyriaConstants.STORAGE_MONETDB_PORT;
+        } else {
+            port = databasePort;
```
dhalperi: Use the [`firstNonNull` function](http://docs.guava-libraries.googlecode.com/git/javadoc/com/google/common/base/MoreObjects.html#firstNonNull(T,%20T)) to simplify this code:

`port = MoreObjects.firstNonNull(databasePort, MyriaConstants.STORAGE_MONETDB_PORT);`

PR #714, opened 2015-03-06
Sending relation: create the index we need

dhalperi:
From 20s of seconds for a few-minutes query to under 1s.


dhalperi:
@domoritz FYI -- since we group by `destWorkerId`, including it in the index lets the database use a streaming instead of a hash-based aggregate and sped up query times dramatically. Not that I understand why, as there are only something like 2100 hash keys... (up to 30 fragments * 72 workers).

domoritz:
Odd but awesome.

Issue #697, opened 2015-02-11
deprecated example json queries


jingjingwang:
There are queries in jsonQueries that are not valid anymore, mostly caused by having "opId"s to be strings not integers. (The first issue came from 544 students! ;)
Also it would be nice to have small text dataset files associated with each example query to help them start.

PR #716, opened 2015-03-10
updating all json queries, mostly changing opid and aggregators

jingjingwang:
All json queries are checked by /query/validate, I think. There is one fix that I didn't know how to do, in lsst_vaspol/lsstQuery_oneIteration.json, there is one predicate called WithinSumRangePredicate which is not supported anymore. If anyone knows how should we replace it please comment.


jingjingwang:
solves #697.

PR #718, opened 2015-03-10
Several extensions

jingjingwang:


dhalperi:
@jingjingwang bump – I would still like you to a test for the bug you fixed, if you don't mind :)

dhalperi:
(When you get back from vacation, of courwe ;)

jingjingwang:
Test added. I gave up the idea of merging this tupleEquals() with the most generic one since it's on the critical path of *CountingJoins and I didn't want to add additional overhead to it. So I simply extended the test data to make the bug appear. Sorry for this 1-month delay!

jingjingwang:
I'm assigning this to tobin since I'm not sure when you're coming back :)

jingjingwang:
I think `&&` always comes before `||`? e.g. http://introcs.cs.princeton.edu/java/11precedence/

senderista:
OK, you're right, but you should absolutely not expect anyone to remember that (I could certainly never remember the order). Expressions with both logical AND and OR should always be disambiguated by parentheses.

src/edu/washington/escience/myria/operator/SymmetricHashJoin.java
```diff
@@ -568,6 +568,11 @@ protected TupleBatch fetchNextReady() throws DbException {
           break;
         }
       } else {
+        if (pollLeft && order.equals(JoinPullOrder.LEFT_EOS) || !pollLeft && order.equals(JoinPullOrder.RIGHT_EOS)) {
```
senderista: Sorry, but parens still need to be added here.

PR #719, opened 2015-03-10
Code/test changes to enable modulo operator.

senderista:
Verifying that floating-point arguments for modulo operator fail typechecking. I didn't create any tests for mixed integral/floating-point types; should I?


dhalperi:
LGTM.

Issue #715, opened 2015-03-06
Specify data partitioning during ingestion


jingjingwang:
Currently a table is always round-robin partitioned, but we may want to specify a particular partitioning strategy when ingesting a table, e.g. hash-partition on some columns, or even having replications. There are several steps in my mind:
1) adding a partition function to JSON
2) adding the information of how a table is partitioned to catalog. We may need to know if a logical table has multiple physical tables that are partitioned in different ways to serve 3)
3) making the optimizer be aware of the information. 

I'll be pushing out 1) soon. A draft version of 2) seems to be not hard. 3) is a RACO issue.

domoritz:
Will you be using shuffle without a hash function and just mod? Then we could just sue apply to achieve custom partitioning.

dhalperi:
I think for the title of this issue, you need a subset of the lines of code in 3d406019576b505071ba77c67a756aa6a90b48e5 .

dhalperi:
Except that, I think, that commit did it wrong. It shouldn't be `String partitionFunction`, it should be `PartitionFunction pf` -- we already can serialize and deserialize partition functions, after all!

PR #720, opened 2015-03-10
Add partition function to ingest

jingjingwang:
Solves #715.


src/edu/washington/escience/myria/coordinator/catalog/MasterCatalog.java
```diff
@@ -1073,8 +1079,16 @@ protected Schema job(final SQLiteConnection sqliteConnection) throws CatalogExce
         long numTuples = statement.columnLong(3);
         long queryId = statement.columnLong(4);
         String created = statement.columnString(5);
+        PartitionFunction howPartitioned;
+        try {
+          howPartitioned =
+              MyriaJsonMapperProvider.getMapper().readValue(statement.columnString(6), PartitionFunction.class);
+        } catch (final IOException e) {
+          LOGGER.debug("Error deserializing howPartitioned for dataset #{}", relationKey.toString(), e);
```
dhalperi: the `toString()` is unnecessary; it just adds work and raises the overhead when debug is not enabled.

src/edu/washington/escience/myria/operator/network/partition/PartitionFunction.java
```diff
@@ -23,6 +23,7 @@
     @Type(value = RoundRobinPartitionFunction.class, name = "RoundRobin"),
     @Type(value = SingleFieldHashPartitionFunction.class, name = "SingleFieldHash"),
     @Type(value = MultiFieldHashPartitionFunction.class, name = "MultiFieldHash"),
+    @Type(value = UnknownPartitionFunction.class, name = "Unknown"),
```
dhalperi: this can't be what you want, can it?

src/edu/washington/escience/myria/api/encoding/DatasetEncoding.java
```diff
@@ -23,4 +25,5 @@
   public Boolean importFromDatabase;
   public List<List<IndexRef>> indexes;
   public Boolean overwrite;
+  public PartitionFunction howPartitioned = new RoundRobinPartitionFunction(null);
```
dhalperi: I'm confused -- isn't this the partition function for ingest? `howPartitioned` does not make sense as a field name.

src/edu/washington/escience/myria/api/encoding/DatasetStatus.java
```diff
@@ -21,16 +22,19 @@
...
    * @param queryId The query that created this dataset.
    * @param created When this dataset was created, in ISO8601 format.
+   * @param howPartitioned How this dataset was partitioned.
    */
   @JsonCreator
   public DatasetStatus(@JsonProperty("relationKey") final RelationKey relationKey,
       @JsonProperty("schema") final Schema schema, @JsonProperty("numTuples") final long numTuples,
-      @JsonProperty("queryId") final long queryId, @JsonProperty("created") final String created) {
+      @JsonProperty("queryId") final long queryId, @JsonProperty("created") final String created,
+      @JsonProperty("howPartitioned") final PartitionFunction howPartitioned) {
```
dhalperi: don't you also need to expose the workers (and the order of the shards) in the API to make this information actionable?

What about other interesting properties -- sorted, min value, max value, unique, key, etc. Rather than adding a specific field for partitioning maybe just an interesting properties dictionary?

src/edu/washington/escience/myria/api/encoding/DatasetEncoding.java
jingjingwang: So maybe just call it partitionFunction? I was trying to use the same name for every relevant field.

src/edu/washington/escience/myria/coordinator/catalog/MasterCatalog.java
jingjingwang: Yes!

src/edu/washington/escience/myria/operator/network/partition/PartitionFunction.java
jingjingwang: I don't get it -- I need a subclass of PartitionFunction to indicate an unknown partition function. Are you saying I should just use a string "unknown" instead of a new class?

src/edu/washington/escience/myria/api/encoding/DatasetStatus.java
jingjingwang: Yeah there could be a lot... will extend it to a dictionary but only add workers for this PR.

src/edu/washington/escience/myria/coordinator/catalog/MasterCatalog.java
jingjingwang: Fixed in 71e2ea3.

src/edu/washington/escience/myria/operator/network/partition/PartitionFunction.java
jingjingwang: Anyway it's now removed in 71e2ea3.

src/edu/washington/escience/myria/api/encoding/DatasetEncoding.java
jingjingwang: Fixed in 71e2ea3.

src/edu/washington/escience/myria/api/encoding/DatasetStatus.java
jingjingwang: So I added a new class HowPartitioned, which can be created/dumped as a JSON object, and put the partiton function & worker array inside. We can extend it to have more info if needed.

jingjingwang:
I'm assigning this to tobin since I'm not sure when you're coming back :)

src/edu/washington/escience/myria/operator/network/partition/HowPartitioned.java
```diff
@@ -0,0 +1,60 @@
...
+ * worker IDs. More information can be added in the future if needed.
+ */
+public class HowPartitioned {
+
+  /** The partition function which was used to partition the dataset. Null means unknown. */
+  @JsonProperty
+  private PartitionFunction pf = null;
+  /** The sequence of workers that the dataset is partitioned on. Null means unknown. */
+  @JsonProperty
+  private int[] workers = null;
```
senderista: Why isn't this a List<Integer> or Set<Integer> to begin with? Isn't the JSON mapping just as simple?

src/edu/washington/escience/myria/operator/network/partition/HowPartitioned.java
```diff
@@ -0,0 +1,60 @@
...
+  public PartitionFunction getPf() {
+    return pf;
+  }
+
+  /** @return the workers as a set. */
+  public Set<Integer> getWorkersAsSet() {
+    if (workers == null) {
+      return null;
+    }
+    return ImmutableSet.copyOf(Ints.asList(workers));
```
senderista: Shouldn't we do some validation here, at least on uniqueness if not order?

src/edu/washington/escience/myria/operator/network/partition/HowPartitioned.java
```diff
@@ -0,0 +1,60 @@
...
+    return new HowPartitioned(pf, workers);
+  }
+
+  /** @return the partition function. */
+  public PartitionFunction getPf() {
+    return pf;
+  }
+
+  /** @return the workers as a set. */
+  public Set<Integer> getWorkersAsSet() {
```
senderista: Is order really irrelevant, or are clients expected to sort the set of worker IDs? Are you only using `Set` to enforce uniqueness? For a List, you could validate both uniqueness and ordering with Guava's `Ordering.natural().isStrictlyOrdered()`.

src/edu/washington/escience/myria/util/JsonAPIUtils.java
```diff
@@ -179,4 +181,24 @@ public static HttpURLConnection ingestData(final String masterHostname, final in
...
     return conn;
   }
+
+  /**
+   * @param masterHostname master hostname
+   * @param apiPort rest api port
+   * @return a HTTPURLConnection instance of retrieving responses.
+   * @throws IOException if IO errors
+   * */
+  public HttpURLConnection getDatasetStatus(final String masterHostname, final int apiPort, final RelationKey key)
```
senderista: Where is this new method used? I see the name used elsewhere but not the method defined in this class.

src/edu/washington/escience/myria/operator/network/partition/HowPartitioned.java
```diff
@@ -0,0 +1,60 @@
...
+
+import com.fasterxml.jackson.annotation.JsonCreator;
+import com.fasterxml.jackson.annotation.JsonProperty;
+import com.google.common.primitives.Ints;
+
+/**
+ * How a dataset is partitioned. There are two fields being recorded for now, the partition function and the sequence of
+ * worker IDs. More information can be added in the future if needed.
+ */
+public class HowPartitioned {
```
senderista: This name seems really odd to me. Why not just something like PartitionInfo or PartitioningInfo?

src/edu/washington/escience/myria/util/JsonAPIUtils.java
jingjingwang: You're right, it's not used. I believe it was called by something that I removed. Will remove it in coming fixes.

src/edu/washington/escience/myria/operator/network/partition/HowPartitioned.java
jingjingwang: Because I was trying to be consistent with the master catalog. This information is stored in `stored_relations` as `how_partitioned`.

src/edu/washington/escience/myria/operator/network/partition/HowPartitioned.java
jingjingwang: I added this function only to be consistent with the current API type requirements, since `DatasetResource.replaceDataset()`, which calls `doIngest()` in the end, needs to pass in a set of workers. Eventually, this set is coverted into a sorted `int[]` in `MyriaUtils.integerSetToIntArray()`.

src/edu/washington/escience/myria/operator/network/partition/HowPartitioned.java
jingjingwang: I guess List is the same as an int[] regarding JSON mapping..? I didn't use Set because JSON doesn't have Set as a data type and I simply want to avoid confusion.

src/edu/washington/escience/myria/operator/network/partition/HowPartitioned.java
senderista: How hard is it to change that attribute name? The name seems less appropriate as partitioning information becomes more detailed than just the general partitioning strategy used.

jingjingwang: I think `how_partitioned` is not just about general partitioning strategy, it depends on what we put inside of it. Currently it has the function and the worker array, we can definitely add more if we want. PartitionInfo, on the other hand, sounds vague to me since "Info" doesn't indicate anything meaningful. 
Changing the name is not hard, but we need to remember to change the master catalog again.

src/edu/washington/escience/myria/coordinator/catalog/MasterCatalog.java
```diff
@@ -1561,6 +1574,48 @@ protected Object job(final SQLiteConnection sqliteConnection) throws CatalogExce
...
+            statement.stepThrough();
+            statement.dispose();
+            statement = null;
+          } catch (final SQLiteException e) {
+            throw new CatalogException(e);
+          }
+          return null;
+        }
+      }).get();
+    } catch (InterruptedException | ExecutionException e) {
```
senderista: You should reset interrupted status when you catch `InterruptedException` since you aren't rethrowing the `InterruptedException` (e.g., http://stackoverflow.com/questions/3976344/handling-interruptedexception-in-java):

`Thread.currentThread().interrupt()`

src/edu/washington/escience/myria/operator/network/partition/HowPartitioned.java
senderista: Could you change to a List then and use ImmutableList internally so we don't have a reference to an externally supplied mutable object?

src/edu/washington/escience/myria/operator/network/partition/HowPartitioned.java
senderista: Actually, since the `List` or `int[]` instance variable `workers` isn't used anywhere, why not just have a single `ImmutableSet` instance variable that we initialize from the `workers` param in the constructor and return in this method? Does `null` have special semantics (it would generally be preferable to have callers expect an empty set rather than `null`)?

src/edu/washington/escience/myria/operator/network/partition/HowPartitioned.java
senderista: OK, I think the name is fine :)

src/edu/washington/escience/myria/coordinator/catalog/MasterCatalog.java
jingjingwang: I think this piece of code is different from the link that you gave. Here, `queue` is a `SQLiteQueue` object which executes jobs in a separated thread. This `InterruptedException` indicates that the SQLite thread was interrupted. The thread running code in `MasterCatalog` shouldn't be in the `interrupted` state. Here, the semantics is: something is wrong when updating the catalog, so just throw a `CatalogException` and return, instead of: the thread executing `updateHowPartitioned()` should stop and do something else -- but up to the caller to decide.

src/edu/washington/escience/myria/operator/network/partition/HowPartitioned.java
jingjingwang: Right, so I copied the `int[]` as an `ImmutableSet`, which also solved the problem above. However, `null` indeed has special semantics, it means the workers are not specified during data ingestion. In this case, all the alive workers are used. See https://github.com/uwescience/myria/blob/master/src/edu/washington/escience/myria/api/DatasetResource.java#L417.

src/edu/washington/escience/myria/coordinator/catalog/MasterCatalog.java
senderista: Right, I missed that the unwrapped exception from the `Future` was thrown on a different thread. It's the task code itself which would need to deal with `InterruptedException` if it called a blocking operation. So never mind :)

Issue #706, opened 2015-03-03
implement a document splitter / string splitter (2 steps)


dhalperi:
1. Build a standalone operator that takes in a string and splits it on a parameter regex. For each input row, e.g. `(x, y, z, s)` where `s` is the string to be split, the output should look like this:

    ```
(x, y, z, s1)
(x, y, z, s2)
...
(x, y, z, sN)
```

    where `s1`... `sN` are the `N` outputs of `s.split()`. This will get you familiar with operators, batches, constructing tuples, etc.

2. Refactor this operator into an abstract operator type (takes in a tuple, produces one or more tuples per row based on some abstract interface) and using `split` as one concrete implementation.

    A second instantation of this operator type would be the `repeat` operator that takes in a tuple `(x,y,z,N)` and emits `N` copies of the tuple `(x,y,z)`.

@mbalazin is this okay to list as a starter task, or did you have someone in mind to work on the string stuff?

domoritz:
I talked with @radion about an operator that takes one row and produces multiple rows. It's really powerful to implement algorithms in myria e.g. for sampling.

mbalazin:
There is a student in 544 who is working on text processing as his class
project. The project will be done in 1.5 weeks. It's not clear how much he
will manage to accomplish since quarters are so short. I don't think he
will implement the operator that you describe. I will check with him.





Magdalena Balazinska
Jean Loup Baer Professor of Computer Science and Engineering
Associate Professor, Dept. of Computer Science & Engineering
Director of the IGERT PhD Program in Big Data / Data Science
Senior Data Science Fellow of the eScience Institute
University of Washington

dhalperi:
List of things we know meet this interface:

- `range(a,b,step)`
- `clone(value, N)`
- `split(string, pattern)`
- `counter(N)`
- @billhowe to fill in: matlab grid cross product
- linear algebra, @ryanmaas to fill in: e.g., @dhalperi's guess `eigenvalues(Matrix)`
- `explode(matrix) -> (i, j , value)`
- and any other "unnest"-y things we think of

PR #722, opened 2015-03-12
New string split operator.

senderista:
Some rather arbitrary choices of convention: column to be split (there can be only one) is specified in operator constructor by name rather than by ordinal, column to be split is deleted from output schema, column containing split results is appended to output schema (with name "_splits" since I couldn't specify a null or empty string). Basic tests added, but more cases might be desirable.

Note that this is just a first pass; the final implementation will likely be something like an expression that is evaluated by a generic multi-row-output apply operator. Also multi-column splits are possible (where we would take cartesian product of all results).


src/edu/washington/escience/myria/operator/Split.java
```diff
@@ -0,0 +1,146 @@
...
+import edu.washington.escience.myria.Schema;
+import edu.washington.escience.myria.Type;
+import edu.washington.escience.myria.column.Column;
+import edu.washington.escience.myria.column.builder.ColumnBuilder;
+import edu.washington.escience.myria.column.builder.ColumnFactory;
+import edu.washington.escience.myria.storage.TupleBatch;
+import edu.washington.escience.myria.storage.TupleUtils;
+
+/**
+ * Generic apply operator.
```
dhalperi: Presumably this comment could be a little better

src/edu/washington/escience/myria/operator/Split.java
```diff
@@ -0,0 +1,146 @@
...
+ */
+public class Split extends UnaryOperator {
+  /***/
+  private static final long serialVersionUID = 1L;
+
+  /**
+   * Name of column to split on regex. Must have String type.
+   */
+  private final String columnName;
```
dhalperi: We tend towards using indices rather than column names for this purpose. The reason is that the optimizer refers to everything by position and there are often columns with unimportant/autogenerated names at intermediate nodes in the query plan.

src/edu/washington/escience/myria/operator/Split.java
```diff
@@ -0,0 +1,146 @@
...
+  public Split(final Operator child, final String columnName, final String regex) {
+    super(child);
+    this.columnName = columnName;
+    pattern = Pattern.compile(regex);
+  }
+
+  @Override
+  protected TupleBatch fetchNextReady() throws DbException {
+    final Operator child = getChild();
+    for (TupleBatch tb = child.nextReady(); tb != null; tb = child.nextReady()) {
```
dhalperi: Why is there a `for` loop here? If we go into the loop once, we will `return` out of it at L102. I think you can simplify to something like:

```java
TupleBatch tb = child.nextReader();
if (tb == null) {
    return null;
}
...
```

src/edu/washington/escience/myria/operator/Split.java
```diff
@@ -0,0 +1,146 @@
...
+      List<ColumnBuilder<?>> colBuilders = new ArrayList<>(tb.numColumns());
```
dhalperi: Can you look at `TupleBatchBuilder` and see if it satisfies your needs? I would have expected you to use it...

src/edu/washington/escience/myria/operator/Split.java
```diff
@@ -0,0 +1,146 @@
...
+            splitColBuilder.appendString(segment);
+            // For each split segment, duplicate the values of all other columns in this row.
+            for (int inputColIdx = 0; inputColIdx < inputColIndexToOutputColIndex.length; ++inputColIdx) {
+              int outputColIdx = inputColIndexToOutputColIndex[inputColIdx];
+              if (outputColIdx != -1) {
+                TupleUtils.copyValue(tb.asColumn(inputColIdx), rowIdx, colBuilders.get(outputColIdx));
+              }
+            }
+          }
+        } catch (Exception e) {
```
dhalperi: We don't do blanket `Exception` in Myria. http://www.javaworld.com/article/2077500/testing-debugging/java-tip-134--when-catching-exceptions--don-t-cast-your-net-too-wide.html Did you have a reason for doing so here?

What are the actual expected exceptions?

src/edu/washington/escience/myria/operator/Split.java
```diff
@@ -0,0 +1,146 @@
...
+        // Offset output column index to compensate for deleted split column index.
+        inputColIndexToOutputColIndex[colIdx] = colIdx - 1;
+        colsToKeep[colIdx - 1] = colIdx;
+      } else {
+        inputColIndexToOutputColIndex[colIdx] = colIdx;
+        colsToKeep[colIdx] = colIdx;
+      }
+    }
+    final Schema subSchema = inputSchema.getSubSchema(colsToKeep);
+    outputSchema = Schema.appendColumn(subSchema, Type.STRING_TYPE, "_splits");
```
dhalperi: the logic to compute `outputSchema` belongs in `generateSchema`... see, e.g, https://github.com/uwescience/myria/blob/master/src/edu/washington/escience/myria/operator/agg/Aggregate.java#L90

src/edu/washington/escience/myria/operator/Split.java
```diff
@@ -0,0 +1,146 @@
...
+  /**
+   * Index of column to split in input schema (and anonymous result column in output schema).
+   */
+  private int splitColIdx;
+
+  /**
+   * Array of column indices in output schema corresponding to column indices of input schema. The column to be split
+   * has its index mapped to the sentinel value -1.
+   */
+  private int[] inputColIndexToOutputColIndex;
```
dhalperi: I think you can simplify this logic. If we define the output of `split` to be like this:

Given input `R(a, b, c)`, `split(R, b)` should return a relation `(a, c, s1), (a,c,s2), ...`

Then we can just make this an array of all the indices that are not split. In this case, it would be `[0,2]` to represent `a` and `c`. Then the code below can ignore any sentinel checks and just prepend the columns listed herein.

src/edu/washington/escience/myria/operator/Split.java
```diff
@@ -0,0 +1,146 @@
...
+  @Override
+  protected void init(final ImmutableMap<String, Object> execEnvVars) throws DbException {
+    final Operator child = getChild();
+    Preconditions.checkNotNull(child, "Child operator cannot be null");
+    final Schema inputSchema = child.getSchema();
+    Preconditions.checkNotNull(inputSchema, "Child schema cannot be null");
+    splitColIdx = inputSchema.columnNameToIndex(columnName);
+    final Type colType = inputSchema.getColumnType(splitColIdx);
+    Preconditions.checkState(colType == Type.STRING_TYPE, "Column to split \"" + columnName
+        + "\" must have type STRING_TYPE but has type " + colType);
```
dhalperi: in all the guava functions like `Preconditions`, it is more efficient to write them as follows:

```java
Preconditions.checkState(colType == Type.STRING_TYPE, "Column to split \"%s\" (%s) must have type STRING_TYPE", columnName, colType);
```

This avoids unnecessary string construction.

senderista:
I now doubt that I made the right decision in deleting the column being split from the input schema and appending it to the output schema instead. At the least, I think it violates the Principle of Least Surprise (and doesn't seem to be how Hive does it with LATERAL VIEW). It also complicates the code considerably for a probably negative usability impact. So I'd propose that I implement the obvious approach of placing the column to be split and the column with split results in the same ordinal position. What do you think?

src/edu/washington/escience/myria/operator/Split.java
senderista: Can't find that in a search of the myria repo...

senderista: Oh, `TupleBatchBuffer`? I recall briefly looking at that, I'll look again...

src/edu/washington/escience/myria/operator/Split.java
senderista: Right, I put it in init() because I assumed each call to generateSchema() (wherever that happens in the framework) should return the same schema (and therefore should be memoized), and also because I happened to need outputSchema in fetchNextReady(), and couldn't be sure that generateSchema() would have been called by that point. If I can assume that, then I could move the schema-generating logic to generateSchema, but I would think I still want to cache the result in an instance field.

src/edu/washington/escience/myria/operator/Split.java
senderista: Agree, I'd forgotten those functions did string formatting

src/edu/washington/escience/myria/operator/Split.java
dhalperi: Please read carefully the documentation on the actual functions that are overriden from `Operator`: https://github.com/uwescience/myria/blob/master/src/edu/washington/escience/myria/operator/Operator.java#L481

In particular, your `generateSchema` does not handle `null` values properly.

You can also look at that code to see how we already handle the memoization cleanly.

dhalperi:
Feel free to keep the original column; if you do this then you should probably put the splits at the end. Then you can simply copy the entire row on its own with no index trickery.

dhalperi:
It would be good to have a test where a single column splits into many values, say, `2*TupleBatch.BATCH_SIZE+1`. (I actually think that such a test may have broken your old code before you picked up `TupleBatchBuffer`.)

src/edu/washington/escience/myria/operator/Split.java
```diff
@@ -0,0 +1,124 @@
...
+import edu.washington.escience.myria.storage.TupleBatchBuffer;
+import edu.washington.escience.myria.storage.TupleUtils;
+
+/**
+ * Operator which splits a string-valued column on a Java regular expression and duplicates the input row with each
+ * segment of the split result.
+ * 
+ * E.g., (1, 2, "foo:bar:baz") -> (1, 2, "foo"), (1, 2, "bar"), (1, 2, "baz")
+ */
+public class Split extends UnaryOperator {
```
dhalperi: do you have `checkstyle` set up per the instructions? I think (but have not verified) that you ought to get a Checkstyle warning about this class not being `final`... which you should.

src/edu/washington/escience/myria/operator/Split.java
```diff
@@ -0,0 +1,124 @@
...
+
+  /**
+   * Compiled regex to split input tuples.
+   */
+  private final Pattern pattern;
+
+  /**
+   * Schema of output tuples.
+   */
+  private Schema outputSchema;
```
dhalperi: not necessary; just use `getSchema()`. Let `Operator` do the work that is common to all operators.

src/edu/washington/escience/myria/operator/Split.java
```diff
@@ -0,0 +1,124 @@
...
+  /**
+   * 
+   * @param child child operator that data is fetched from
+   * @param columnName name of string column to split using regex
+   * @param regex regular expression to split string input column on
+   */
+  public Split(final Operator child, final int splitColumnIndex, final String regex) {
+    super(child);
+    this.splitColumnIndex = splitColumnIndex;
+    pattern = Pattern.compile(regex);
```
dhalperi: what happens if `regex` is `null`?

src/edu/washington/escience/myria/operator/Split.java
```diff
@@ -0,0 +1,124 @@
...
+   */
+  private TupleBatchBuffer outputBuffer;
+
+  /**
+   * 
+   * @param child child operator that data is fetched from
+   * @param columnName name of string column to split using regex
+   * @param regex regular expression to split string input column on
+   */
+  public Split(final Operator child, final int splitColumnIndex, final String regex) {
```
dhalperi: I am trying to push the maximal use of JSR-305 `Nullable` and `NotNull` etc. annotations for new code, and to add them to old code when I touch it. Can these parameters be `null`, and add the relevant notations?

src/edu/washington/escience/myria/operator/Split.java
```diff
@@ -0,0 +1,124 @@
...
+    super(child);
+    this.splitColumnIndex = splitColumnIndex;
+    pattern = Pattern.compile(regex);
+  }
+
+  private TupleBatch getChildBatch() throws DbException {
+    TupleBatch childTuples;
+    try {
+      childTuples = getChild().fetchNextReady();
```
dhalperi: use `nextReady()`, not `fetchNextReady()`. There are some semantic issues that should be documented in `Operator` and, if not, @jingjingwang can explain

src/edu/washington/escience/myria/operator/Split.java
```diff
@@ -0,0 +1,124 @@
...
+    } catch (Exception e) {
+      // only wrap checked exceptions that do not extend DbException
+      Throwables.propagateIfPossible(e, DbException.class);
+      throw new DbException(e);
+    }
+    return childTuples;
+  }
```
dhalperi: can you say more about what's going on here / what your logic is? You may know something that I don't....

From my belief: `getChildBatch()` can throw a `DbException` only (in `nextReady()`). So you shouldn't need to do any exception-handling here at all.

dhalperi: (well, or `RuntimeException` but we don't want those to be wrapped in `DbException` anyway)

src/edu/washington/escience/myria/operator/Split.java
dhalperi: Just created #723

src/edu/washington/escience/myria/operator/Split.java
```diff
@@ -0,0 +1,124 @@
...
+
+  @Override
+  protected TupleBatch fetchNextReady() throws DbException {
+    // If there's a batch already finished, return it, otherwise keep reading
+    // batches from the child until we have a full batch or the child returns null.
+    TupleBatch inputTuples;
+    while (!outputBuffer.hasFilledTB() && (inputTuples = getChildBatch()) != null) {
```
dhalperi: I know it adds 2 lines of code, but I've been trying to discourage the inline assignments. It's easy to miss them when reading the code, so makes it much easier (imo) to understand what's going on.

src/edu/washington/escience/myria/operator/Split.java
```diff
@@ -0,0 +1,124 @@
...
+
+    outputBuffer = new TupleBatchBuffer(generateSchema());
+  }
+
+  @Override
+  public Schema generateSchema() {
+    final Schema inputSchema = getChild().getSchema();
+    final List<String> outputColumnNames = new ArrayList<>(inputSchema.getColumnNames());
+    final String splitColumnName = outputColumnNames.get(splitColumnIndex);
+    final String splitResultsColumnName = splitColumnName + "_splits";
```
dhalperi: name choice LGTM

test/edu/washington/escience/myria/operator/SplitTest.java
```diff
@@ -0,0 +1,115 @@
...
+        for (int batchIdx = 0; batchIdx < result.numTuples(); ++batchIdx, ++rowIdx) {
+          assertEquals(expectedResults[rowIdx], result.getString(0, batchIdx));
+        }
+      }
+    }
+    assertEquals(expectedResults.length, rowIdx);
+    splitOp.close();
+  }
+
+  @Test(expected = Exception.class)
```
dhalperi: what type of `Exception` is expected? Wouldn't want the test to succeed because of, say, an `NPE` introduced somewhere by accident.

test/edu/washington/escience/myria/operator/SplitTest.java
```diff
@@ -0,0 +1,115 @@
...
+  public void testSplitColumnInvalidType() throws DbException {
+    final Schema schema = new Schema(ImmutableList.of(Type.LONG_TYPE), ImmutableList.of("long"));
+    final TupleBatchBuffer input = new TupleBatchBuffer(schema);
+    input.putLong(0, 1L);
+    Split splitOp = new Split(new TupleSource(input), 0, ":");
+    splitOp.open(TestEnvVars.get());
+    splitOp.close();
+  }
+
+  @Test(expected = Exception.class)
```
dhalperi: what type of `Exception` is expected? Wouldn't want the test to succeed because of, say, an `NPE` introduced somewhere by accident.

dhalperi:
Also, and I'll add this to the getting started instructions, just read through Google's Java style guide once.

https://google-styleguide.googlecode.com/svn/trunk/javaguide.html

src/edu/washington/escience/myria/operator/Split.java
senderista: Assume PatternSyntaxException would be thrown, but added precondition check.

src/edu/washington/escience/myria/operator/Split.java
senderista: I do not get a checkstyle warning for this. There are other classes extending UnaryOperator which are not final (because they're extended by other classes, e.g., Apply).

src/edu/washington/escience/myria/operator/Split.java
senderista: I think this is a fantastic idea, as long we're forced to use nullable references and not proper Option types :)

src/edu/washington/escience/myria/operator/Split.java
senderista: Right, calling `nextReady()` instead of `fetchNextReady()` avoids any need for explicit exception handling (and for this helper method).

src/edu/washington/escience/myria/operator/Split.java
senderista: I hate exiting early from loops, but a `break` looks like the simplest way to avoid inline assignment.

test/edu/washington/escience/myria/operator/SplitTest.java
senderista: I really don't like coupling unit tests to an exception type that could easily change with a trivial change in validation logic, but I see your point.

senderista:
FYI, I found an Eclipse format config for the Google style guide here: https://code.google.com/p/google-styleguide/source/browse/trunk/eclipse-java-google-style.xml

senderista:
Now I'm wondering if I need to check for `child.eos()` (and possibly `child.eoi()` as well, I'm unclear on the semantics) in the while-loop condition in `fetchNextReady()`. If so, is it possible for the child to return `null` from `nextReady()` on one call but non-`null` on a subsequent call (assuming `eos()` isn't true)? If so, then I shouldn't be breaking out of the loop on `null`.

senderista:
edu.washington.escience.myria.systemtest.JsonQuerySubmitTest.ingestTest is failing for reasons unrelated to this PR

src/edu/washington/escience/myria/operator/Operator.java
```diff
@@ -67,7 +67,7 @@
   private Schema schema;
 
   /**
-   * EOS. Initially set it as true;
+   * End of stream. Initially set it as true;
```
dhalperi: Since we use the acronym so much, may as well add `End of stream (EOS)`.

src/edu/washington/escience/myria/operator/Split.java
```diff
@@ -0,0 +1,111 @@
...
+
+  /**
+   * 
+   * @param child child operator that data is fetched from
+   * @param splitColumnIndex index of string column to split using {@link #regex}
+   * @param regex regular expression to split value of column at {@link #splitColumnIndex}
+   */
+  public Split(@Nonnull final Operator child, final int splitColumnIndex, @Nonnull final String regex) {
+    super(child);
+    Preconditions.checkNotNull(child, "Child operator cannot be null");
```
dhalperi: for these, just put the name of the `null` parameter – the message will be `NullPointerException: child` and the extra text is just a waste of your time and bytes :)

src/edu/washington/escience/myria/operator/Split.java
```diff
@@ -0,0 +1,111 @@
...
+      for (int rowIdx = 0; rowIdx < inputTuples.numTuples(); ++rowIdx) {
+        String colValue = inputTuples.getString(splitColumnIndex, rowIdx);
+        // We must specify a negative value for the limit parameter to avoid discarding trailing empty strings:
+        // http://docs.oracle.com/javase/7/docs/api/java/lang/String.html#split(java.lang.String,%20int)
+        String[] splits = pattern.split(colValue, -1);
+        for (String segment : splits) {
+          outputBuffer.putString(splitColumnIndex, segment);
+          // For each split segment, duplicate the values of all other columns in this row.
+          for (int colIdx = 0; colIdx < inputTuples.numColumns(); ++colIdx) {
+            if (colIdx != splitColumnIndex) {
```
dhalperi: I thought you were going to keep the original column and append the splits?

Regardless, the splits should be at the end. That's the normal user expectation – existing columns followed by new columns.

test/edu/washington/escience/myria/operator/SplitTest.java
```diff
@@ -0,0 +1,167 @@
...
+
+  @Test
+  public void testGeneratedSplits() throws DbException {
+    final Object[][] expectedResults =
+        { {true, "foo", 1L, 0.1}, {true, "bar", 1L, 0.1}, {true, "baz", 1L, 0.1,},
+            {false, "", 2L, 0.2}, {false, "qux", 2L, 0.2}, {false, "", 2L, 0.2},
+            {false, "", 2L, 0.2}};
+    final Schema schema =
+        new Schema(ImmutableList.of(Type.BOOLEAN_TYPE, Type.STRING_TYPE, Type.LONG_TYPE,
+            Type.DOUBLE_TYPE), ImmutableList.of("bool", "string", "long", "double"));
```
dhalperi: FYI, you're welcome to use `Schema.ofFields` – I find it the most clear way to define a Schema in practice.

```java
Schema.ofFields("bool", Type.BOOLEAN_TYPE, "string", Type.STRING_TYPE, ...)
```

test/edu/washington/escience/myria/operator/SplitTest.java
```diff
@@ -0,0 +1,167 @@
...
+    Split splitOp = new Split(new TupleSource(input), 1, ":");
+
+    splitOp.open(TestEnvVars.get());
+    TupleBatch result;
+    int rowIdx = 0;
+    while (!splitOp.eos()) {
+      result = splitOp.nextReady();
+      if (result != null) {
+        assertEquals(schema.numColumns(), result.getSchema().numColumns());
+        assertEquals(Type.BOOLEAN_TYPE, result.getSchema().getColumnType(0));
```
dhalperi: why not just assert that `schema.getColumnTypes().equals(resultSchema.getColumnTypes())`?

test/edu/washington/escience/myria/operator/SplitTest.java
```diff
@@ -0,0 +1,167 @@
...
+   */
+  @Test
+  public void testAllBatchesReturned() throws DbException {
+
+    final Schema schema =
+        new Schema(ImmutableList.of(Type.STRING_TYPE), ImmutableList.of("joined_ints"));
+    final TupleBatchBuffer input = new TupleBatchBuffer(schema);
+    final long expectedResults = 2 * TupleBatch.BATCH_SIZE + 1;
+    final int intsInRow = 10;
+    for (long i = 0; i < (expectedResults / intsInRow) + 1; ++i) {
```
dhalperi: what's going on here? This seems awfully complicated. Is there a reason you didn't just use a `StringBuilder` and concatenate the ints `0..expectedResults-1`?

```java
StringBuilder sb = new StringBuilder();
for (int i = 0; i < expectedResults; ++i) {
   sb.append(i);
}
input.putString(0, sb.toString());
```

[or something like that; you may need to convert the int to string.]

src/edu/washington/escience/myria/operator/Split.java
senderista: Yes, I now think you're right. The Hive LATERAL VIEW syntax requires you to specify a column alias for the "exploded" column (because it's in the FROM clause), and does not delete the original column from the returned relation's schema.

test/edu/washington/escience/myria/operator/SplitTest.java
senderista: That's better than the parallel array approach, but still too fast and loose for my taste since the names and types can be interleaved arbitrarily :) I don't see why the API doesn't just take a set of name-type pairs--that's both clearer (to me) and safer.

test/edu/washington/escience/myria/operator/SplitTest.java
```diff
@@ -0,0 +1,167 @@
...
+  }
+
+  @Test(expected = IllegalStateException.class)
+  public void testSplitColumnInvalidType() throws DbException {
+    final Schema schema = new Schema(ImmutableList.of(Type.LONG_TYPE), ImmutableList.of("long"));
+    final TupleBatchBuffer input = new TupleBatchBuffer(schema);
+    input.putLong(0, 1L);
+    Split splitOp = new Split(new TupleSource(input), 0, ":");
+    splitOp.open(TestEnvVars.get());
+    splitOp.close();
```
dhalperi: hopefully the `close` isn't necessary?

test/edu/washington/escience/myria/operator/SplitTest.java
dhalperi: Java doesn't provide a reasonable syntax for that.

dhalperi:
returning `null` from `fetchNextReady` means that there can be no output produced right now. If there is new data available later (e.g., it comes in from the network) then yes you can definitely return `null` then return non-null later. I'll add another note to #723

test/edu/washington/escience/myria/operator/SplitTest.java
senderista: I was concerned I would run into a row size limit in Myria (this would be 20001 ints in one row, a few bytes each in ASCII, with an extra byte for each comma).

dhalperi: Presumably that's a case we need to be able to handle.... but if you want,
you should feel free to `%10` or something.


----
Daniel Halperin
Director of Research for Scalable Data Analytics
eScience Institute
University of Washington

senderista:
So the `while` loop calling `child.nextReady()` should just check for `child.eos()` and not `child.eoi()`?

test/edu/washington/escience/myria/operator/SplitTest.java
senderista: Of course it won't be reached after the expected exception is thrown, but maybe for style at least I should put the open/close in a try-finally block?

dhalperi: When I did this, I assumed (perhaps erroneously) that the line I expect to throw the exception should be the last line of the test. That way if something goes wrong with the test I find out.

This is extraordinarily unlikely, but I wouldn't want the test to "pass" because `close` throws the exception.

IMO style and symmetry aren't so important for tests, so much as conciseness/understandability and most importantly, correctness.

dhalperi:
I don't think the `while` loop should care about what state the child is in. If the child has reached one of these `eos/eoi` states, it will return `null`. When it returns `null`, your code will do the right thing. Won't it?

senderista:
OK, so you're saying that if one of my consumers wants to know if I will ever return any more data, they can just call `eos()` on me, and that will return the child's value of `eos()`? So I shouldn't need to care whether my child will ever return more data, as long as my consumer can find that out for themselves? (If the child returns `null`, I will either return `null` or whatever is in the current TupleBatch.)

dhalperi:
As I suggested before, please read through `Operator` to see how `eos` and `eoi` are triggered.

Can I rephrase your last sentence as follows?

> If the child returns null, I will return whatever is in the current TupleBatchBuffer (including null if it is empty).

test/edu/washington/escience/myria/operator/SplitTest.java
senderista: Fair enough.

dhalperi:
Also, please read through random other operators to see how they're implemented. Stay away from complex things like joins, consumers, and anything related to EOS or IDB for now...

maybe look at Filter, Apply, Counter, CrossWithSingleton?

senderista:
Yes, your paraphrase is clearer than my original sentence.

My understanding is that an operator returns EOS if all of its children return EOS, and EOI if any of its children return EOI. Also a child's EOI state is reset whenever it is read, but I still don't understand why (I need an intuitive explanation of what these states *mean*)...

I am probably getting confused from reading more complicated operators like Merge...

dhalperi:
yeah... and even in Merge, it's only worrying about EOS as a way of streamlining which children it looks at. (Children that will never produce any more data never need to be looked at again ;). It's not actually doing anything with the EOS semantics.

test/edu/washington/escience/myria/operator/SplitTest.java
senderista: Well, putting them all into a single row seems to work, so I'll leave it at that.

dhalperi:
So the next thing to do when adding a new operator is make sure that it has a tested JSON encoding.

Because of annoying crap (the JSON libraries have trouble de-serializing cyclic graphs from JSON), we cannot currently deserialize `Operator`s directly. Instead, we add a very small Java class extending `OperatorEncoding` that explicitly defines the (optional and `@Required`) fields for the encoding and constructs the operator after the entire plan is deserialized. Take a look at:

* https://github.com/uwescience/myria/blob/master/src/edu/washington/escience/myria/api/encoding/FilterEncoding.java
* and notice https://github.com/uwescience/myria/blob/212a7e3f6559bda759493775fe0000712165ce89/src/edu/washington/escience/myria/api/encoding/OperatorEncoding.java#L36

After implementing the analogous encoding for `Split`, add a test to ensure that it can be serialized and deserialized and submitted to a query. Currently we're doing this as such: https://github.com/uwescience/myria/blob/master/systemtest/edu/washington/escience/myria/systemtest/JsonOperatorTests.java and sadly, most operators are not tested here, though many of the key operators are tested in the JsonQuerySubmitTests instead.

senderista:
For the JSON encoding test, it seems I need a JSON-serializable leaf operator which yields my test data. TupleSource looks like what I want, but there's no encoding class for it, and I don't see any encoding-compatible class which appears to do anything similar (or is capable of serializing any test data in a JSON query). What's the right approach?

dhalperi:
The usual thing is to ingest some data before the query, and use [`TableScan`](https://github.com/uwescience/myria/blob/master/src/edu/washington/escience/myria/api/encoding/TableScanEncoding.java)

senderista:
I used FileScan instead to avoid ingesting data.

dhalperi:
Why not a `ByteSource`, e.g.?

Hmmmm... Heck, if you use that you don't even need to make it a system test – you ought to be able to test that serializing and deserializing works without starting up the server.

senderista:
Well, I have a lot more confidence in integration tests than in unit tests, although you have to decide what's worth testing at the system level.

dhalperi:
fine w me...


----
Daniel Halperin
Director of Research for Scalable Data Analytics
eScience Institute
University of Washington

dhalperi:
There seem to be some wonky style issues with this PR. E.g., I get 4 new Checkstyle warnings relating to missing spaces, and every time I save a file it reformats the whole thing.

You must not be using the eclipse configuration that is checked into the repository?

If you wish to reformat the files and apply new settings, please:

1. Do it in a separate branch that has no functional changes, and reformat ALL source files in the repository.
2. Make sure that anyone who clones the repository (or pulls from master) and edits a file will automatically use the same style settings.

But in general, I don't understand why you decided to deviate from the checked-in code style.

For fun: http://thecodelesscode.com/case/94?topic=refactoring  (omg this is my favorite weekly blog there is)

dhalperi:
The basic principle is that, when code reviewing, I should be able to tell what are functional and what are stylistic changes; right now there's a large fraction of the diff that's pure style.

Ideally the purely stylistic changes would be encapsulated so that they "don't need" review.

senderista:
Agree 100% on convention, I had installed the Google Java style guide's official Eclipse formatter, which seems to differ from the one checked in, and didn't pay attention to the reformatting noise in the diffs. My bad--I'll revert the formatting changes and use the checked-in formatter. Was the currently checked-in formatter originally derived from the Google style guide, or from your/others' preferences?

dhalperi:
I have a vague impression that it might have been one claimed to be derived from the Google guide, but I think Google didn't have an official one online at the time. Or something like that – no idea really. It could also have been Eclipse's default mildly tweaked.

senderista:
I would say that in at least one concrete case I prefer the Google guide's formatting: in /myria/src/edu/washington/escience/myria/api/encoding/OperatorEncoding.java we have multiple annotations on the same line, which I find quite confusing given that they're deliberately sorted to be easier to vertically scan. Do you have anything against standardizing on the Google guide's formatter and checking in a wholesale reformatting change (in a separate branch/PR of course)?

dhalperi:
Link?

senderista:
E.g., see https://github.com/uwescience/myria/blob/3cdede0153abfd2ae6c704d7ec4995c8cf4e626a/src/edu/washington/escience/myria/api/encoding/OperatorEncoding.java#L45

dhalperi:
oh yeah that annoys the heck out of me, too. I distinctly remember looking for that style change, and failing.

senderista:
I've reverted the formatting changes but will be squashing all these commits onto a new branch (since I mistakenly recycled an old one), and will submit a new PR, so you don't have to deal with the noise.

dhalperi:
don't bother – I mainly use the diff view anyway

dhalperi:
https://github.com/uwescience/myria/pull/722/files

senderista:
OK, I did just push a single squashed commit on the new branch split_operator if you care :)

senderista:
OK to merge?

domoritz:
Cool stuff!

senderista:
Fixes #706 (partially).

Issue #721, opened 2015-03-11
Error when using http/https scheme with UriSource


BrandonHaynes:
Consider the plan below, which sources data from a single URL.  

Upon execution, Myria reports `java.io.IOException: No FileSystem for scheme: http` (full exception below).  I'm virtually positive that I have successfully loaded data using an http scheme (the only other possibility is that I was using a custom `DataSource` implementation).  It's also possible that I had frameworks that expose a `HttpFileSystem` co-installed at the time (e.g., Hadoop's HttpFS).

Is this a regression bug, a configuration issue, or does the UriSource really not currently support these schemes?  It's also possible that I've gone insane, because I was sure that this used to work.

(Originally observed by @ryanmaas) 

```json
{
   "fragments":[
      {
         "workers":[
            1
         ],
         "operators":[
            {
               "source":{
                  "dataType":"URI",
                  "uri":"http://s3-us-west-2.amazonaws.com/myria-sdss/oceanography/smallsample_ocean_seq.csv"
               },
               "opId":0,
               "opType":"FileScan",
               "schema":{
                  "columnNames":[
                     "col",
                     "col2",
                     "col3",
                     "col4",
                     "col5"
                  ],
                  "columnTypes":[
                     "LONG_TYPE",
                     "DOUBLE_TYPE",
                     "DOUBLE_TYPE",
                     "DOUBLE_TYPE",
                     "DOUBLE_TYPE"
                  ]
               }
            },
            {
               "argOverwriteTable":true,
               "opId":1,
               "opType":"DbInsert",
               "relationKey":{
                  "userName":"public",
                  "relationName":"testMyriaUpload",
                  "programName":"adhoc"
               },
               "argChild":0
            }
         ]
      }
   ],
   "logicalRa":"Parallel Import [(1, 'http://s3-us-west-2.amazonaws.com/myria-sdss/oceanography/smallsample_ocean_seq.csv')]",
   "rawQuery":"Parallel Import [(1, 'http://s3-us-west-2.amazonaws.com/myria-sdss/oceanography/smallsample_ocean_seq.csv')]"
}
```

Full exception:

```
edu.washington.escience.myria.DbException: Query #66690.0 failed: java.io.IOException: No FileSystem for scheme: http\n\tat edu.washington.escience.myria.parallel.MasterSubQuery$WorkerExecutionInfo$2.operationComplete(MasterSubQuery.java:105)
[...]

PR #727, opened 2015-03-21
Support http and https schemes in UriSource

BrandonHaynes:
Since there is no registered Hadoop FileSystem for http/https schemes, explicitly check for these cases and use `java.net.URL` methods to retrieve an `InputStream`.

Corrects #721.

### DIFF
--- a/src/edu/washington/escience/myria/io/UriSource.java
+++ b/src/edu/washington/escience/myria/io/UriSource.java
@@ -6,6 +6,7 @@ import java.io.InputStream;
 import java.io.SequenceInputStream;
 import java.io.Serializable;
 import java.net.URI;
+import java.net.URL;
 import java.util.ArrayList;
 import java.util.List;
 import java.util.Objects;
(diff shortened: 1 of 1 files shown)


src/edu/washington/escience/myria/io/UriSource.java
```diff
@@ -51,14 +52,24 @@ public UriSource(@JsonProperty(value = "uri", required = true) final String uri)
 
   @Override
   public InputStream getInputStream() throws IOException {
-    // Use Hadoop's URI parsing machinery to extract an input stream for the underlying URI
+    URI parsedUri = URI.create(uri);
+
+    return parsedUri.getScheme().equals("http") || parsedUri.getScheme().equals("https")
```
dhalperi: I tend to avoid trigraphs unless they're absolutely necessary, but that's up to you.

However, please `()` around the condition of `||` here... the code is correct but you have to know non-obvious parts of [the precedence chart (13 vs 14)](http://introcs.cs.princeton.edu/java/11precedence/).

LGTM!

PR #733, opened 2015-04-08
better error message for TupleBatch constructor

radion:
Error message confused me when I first ran into it.

### DIFF
--- a/src/edu/washington/escience/myria/storage/TupleBatch.java
+++ b/src/edu/washington/escience/myria/storage/TupleBatch.java
@@ -101,8 +101,11 @@ public class TupleBatch implements ReadableTable, Serializable {
     this.columns = ImmutableList.copyOf(Objects.requireNonNull(columns, "columns"));
     Preconditions.checkArgument(columns.size() == schema.numColumns(),
         "Number of columns in data must equal the number of fields in schema");
-    for (Column<?> column : columns) {
-      Preconditions.checkArgument(numTuples == column.size(), "Column %s != %s tuples", column.size(), numTuples);
+    for (int i = 0; i < columns.size(); i++) {
+      Column<?> column = columns.get(i);
+      Preconditions.checkArgument(numTuples == column.size(),
+              "Incorrect size for column %s. Expected %s tuples, but found %s tuples.",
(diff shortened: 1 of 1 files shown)


src/edu/washington/escience/myria/storage/TupleBatch.java
```diff
@@ -101,8 +101,11 @@ public TupleBatch(final Schema schema, final List<? extends Column<?>> columns,
     this.columns = ImmutableList.copyOf(Objects.requireNonNull(columns, "columns"));
     Preconditions.checkArgument(columns.size() == schema.numColumns(),
         "Number of columns in data must equal the number of fields in schema");
-    for (Column<?> column : columns) {
-      Preconditions.checkArgument(numTuples == column.size(), "Column %s != %s tuples", column.size(), numTuples);
+    for (int i = 0; i < columns.size(); i++) {
+      Column<?> column = columns.get(i);
+      Preconditions.checkArgument(numTuples == column.size(),
+              "Incorrect size for column %s. Expected %s tuples, but found %s tuples.",
+              i, numTuples, column.size());
```
domoritz: I is the row, right? Something is wrong here, imho.

radion: I'm looking into why this change broke the build...
I'll rename the variable to 'j' to not confuse with the typical row index variable 'i'

radion:
Should be good now.

src/edu/washington/escience/myria/storage/TupleBatch.java
jingjingwang: I is the column index, not the row. What you have here is correct, the test failed not because of this change. I refreshed it and it passed. 
So I'll revert your following commit. I think "i" is definitely an appropriate name for column index variable :)

jingjingwang: Actually could you do the reversion? I'm not sure how to checkout to your branch.

radion: Done.

jingjingwang:
Cool, thanks!

PR #734, opened 2015-04-11
Utility to create deployment files from the command line

BrandonHaynes:
I need a command-line utility to generate deployment files for EC2- and HPC-oriented deployments.  Since this seems broadly useful, I thought we might include it in the distribution.

### DIFF
--- a/myriadeploy/create_deployment.py
+++ b/myriadeploy/create_deployment.py
@@ -1,0 +1,119 @@
+#!/usr/bin/env python
+
+""" Create a Myria deployment file """
+
+import sys
+import argparse
+from itertools import groupby
+
+
+def get_deployment(path, coordinator_hostname, worker_hostnames, name='myria',
+                   rest_port=8753, database_type='postgresql',
+                   database_port=5432, heap='2g', debug=False,
+                   database_username=None, database_password=None,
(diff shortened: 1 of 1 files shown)


jingjingwang:
If we need to specify different working directories/database names for workers, like what we have in the production, then this seems to be not enough. It would be great if this script can be extend (although there is probably no neat way...). I'm not sure if this functionality is in your plan..?

BrandonHaynes:
The immediate use case here is to enable easier creating of configuration files for HPC deployment, where the set of nodes is not known until job submission time.  It's also really useful for the EC2 side, where this would allow me to avoid having to hardcode the deployment creation process there -- this is brittle, and will break sooner or later.

Extending the utility to support the last few configuration options is totally on the table, although this would increase the complexity of the arguments a bit.  Is it possible to handle that as a follow-on feature, since I need this for HPC scheduler init scripts now?

jingjingwang:
Yes, definitely. I'm just wondering if you plan to do it in this PR or a following PR. Could you add a usage message to it? Basically saying it's not ready if we need to specify database names/working directories.

BrandonHaynes:
Usage is displayed when you execute the utility with no arguments.  I'll look at supporting the full set of deployment options; I agree this is reasonable.

PR #735, opened 2015-04-13
Sampling Operators

radion:
Unit tests forthcoming.


src/edu/washington/escience/myria/operator/SampleWoR.java
```diff
@@ -0,0 +1,66 @@
...
+      // Cannot sampleWoR more tuples than there are.
+      Preconditions.checkState(sampleSize <= populationSize,
+          "Cannot SampleWoR %s tuples from a population of size %s",
+          sampleSize, populationSize);
+
+      getLeft().close();
+      computedSamplingInfo = true;
+    }
+
+    // SampleWoR from the right operator.
```
senderista: This algorithm is standard, but is linear in the population size rather than the sample size. One algorithm for sampling WoR that is linear in the sample size is Floyd's algorithm from _Programming Pearls_ (`N` is population size, `M` is sample size):
```
initialize set S to empty
for J := N-M + 1 to N do
  T := RandInt(1, J)
  if T is not in S then
    insert T in S
  else
    insert J in S
```
This is tricky but can be proven with a simple induction (which just took me an embarrassingly long time to prove).

Anyway, not sure if this is a valid performance concern at all, but just thought I'd mention it. I also wonder how this works for populations whose size isn't known (e.g., streaming intermediate results). Is that a valid scenario? If so, I guess some variant of reservoir sampling would be called for...

src/edu/washington/escience/myria/operator/SampleWR.java
```diff
@@ -0,0 +1,110 @@
...
+   * Generates a sorted sequence of random indices that will be chosen to sample
+   * from.
+   * 
+   * @param populationSize
+   *          size of the population that will be sampled from.
+   * @param sampleSize
+   *          number of samples to draw from the population.
+   * @return a sorted array of indices.
+   */
+  private int[] generateTargetSampleIndices(int populationSize, int sampleSize) {
```
senderista: If you wanted to do reservoir sampling with replacement for sample size `k`, you could run `k` instances of the  reservoir sampling algorithm with `k=1` in parallel.

radion: How would you run that many instances on a small number of worker nodes? What are the performance implications of having separate instances?

src/edu/washington/escience/myria/operator/SampleWoR.java
radion: The assumption here is that we know the population size N in advance. Raco will do this by running a COUNT(*) before sampling. There is overhead in this, but allows for more efficient sampling. Interesting to think about the differences in taking this overhead vs. using reservoir sampling on an unknown stream size.

src/edu/washington/escience/myria/operator/SampleWR.java
senderista: Sorry I wasn't clear, I meant logically parallel, in a single thread. I.e., for sample size `k`, you'd have `k` separate reservoirs of 1 element each, and you would evaluate the 1-element reservoir sampling algorithm independently `k` times for each tuple. Granted, having to generate `k` random numbers for each tuple might be a lot of overhead for large `k`...

src/edu/washington/escience/myria/operator/SampleWoR.java
radion: Concerning the Floyd sampling WoR algorithm, would this be possible when we process only one TupleBatch at a time?

senderista: It is an interesting tradeoff: being able to emit tuples incrementally by evaluating all indices upfront, vs. being able to consume tuples incrementally (without knowing cardinality) by delaying emitting tuples until all have been consumed. Also, of course, maintaining the set of sampled indices takes a lot less space than maintaining the set of sampled tuples.

Are you suggesting using the `COUNT(*)` approach with intermediate results of unknown cardinality? It seems to me that reservoir sampling would be a requirement for using the sampling operators as selection operators anywhere in a query tree...

senderista: The Floyd algorithm can be used to generate a set of random indices before you process any tuples, just like your sampling WR algorithm.

radion:
Update: there are performance implications for TempInserting the entire output of an expression. I'm working on making this more efficient by having each worker TempInsert only a sample of the data.

radion:
The travis fails due to a test that is irrelevant to this PR. I think that the failure will probably be resolved if we just rerun the travis build.

radion:
I think I messed up a rebase here - going to start a new PR for my sampling operators

PR #736, opened 2015-04-21
Flatmap apply

senderista:
The first two commits on this branch can be ignored; they were a prototype to allow me to test vector-valued expressions and evaluators without implementing a new Apply operator. The new interface for vector-valued expressions assumes that they return an Iterable-valued result, which should lazily materialize state whenever possible. The cartesian product implementation in the FlatteningApply operator exploits the Iterable type of expressions to avoid materializing not just the entire product, but the Iterators themselves. The new parallel inheritance hierarchy for flattening evaluators could probably use a refactoring at some point. Also, a FlatteningStatefulApply operator has yet to be implemented.


jingjingwang:
Cool, I think the general idea of this PR is pretty clear, we just need some refactoring to be column-oriented :)

senderista:
@jingjingwang could you comment on "Files changed" rather than "Commits"? I find that much easier to follow, since any individual commit may have little or no relation to the current state of the PR.

senderista:
Fixes #706.

senderista:
@jingjingwang I refactored as we discussed and rebased on latest master; could you review so we can hopefully merge by end of this week?

src/edu/washington/escience/myria/operator/Apply.java
```diff
@@ -110,22 +226,14 @@ protected void init(final ImmutableMap<String, Object> execEnvVars) throws DbExc
...
+      evalBuilder.add(evaluator);
     }
-  }
-
-  /**
-   * @param evaluators the evaluators to set
-   */
-  public void setEvaluators(final ArrayList<GenericEvaluator> evaluators) {
-    emitEvaluators = evaluators;
+    emitEvaluators = evalBuilder.build();
```
jingjingwang: I think here we can call `setEmitEvaluators()` instead.

src/edu/washington/escience/myria/operator/Apply.java
```diff
@@ -47,47 +57,152 @@
...
+   * @param evaluators the evaluators to set
+   */
+  public void setEmitEvaluators(final List<GenericEvaluator> evaluators) {
+    emitEvaluators = ImmutableList.copyOf(evaluators);
+  }
+
+  /**
+   * @return if there are no multivalued emit expressions
+   */
+  private boolean onlySingleValuedExpressions() {
```
jingjingwang: This can be set during initialization as a constant, probably in `setEmitExpressions()`.

src/edu/washington/escience/myria/operator/Apply.java
```diff
@@ -47,47 +57,152 @@
...
+   * @param upperBounds an immutable array of elements representing the sets we are forming the Cartesian product of,
+   *          where each set is of the form [0, i), where i is an element of {@link upperBounds}
+   * @param iteratorIndexes a mutable array of elements representing the current element of the Cartesian product
+   * @return if we have exhausted all elements of the Cartesian product
+   */
+  private boolean computeNextCombination(final int[] upperBounds, final int[] iteratorIndexes) {
+    boolean endOfIteration = false;
+    int lastIteratorPos = iteratorIndexes.length - 1;
+    // Count backward from the innermost iterator to the outermost.
+    for (int iteratorPos = lastIteratorPos; iteratorPos >= 0; --iteratorPos) {
```
jingjingwang: There is no need to loop backward, it's a Cartesian product anyway.

src/edu/washington/escience/myria/expression/CounterExpression.java
```diff
@@ -0,0 +1,76 @@
...
+import com.google.common.base.Preconditions;
+import com.google.common.collect.ImmutableList;
+
+import edu.washington.escience.myria.Type;
+import edu.washington.escience.myria.expression.evaluate.ExpressionOperatorParameter;
+
+/**
+ * Return a sequence of numbers from 0 (inclusive) to operand (exclusive).
+ */
+public class CounterExpression extends UnaryExpression {
```
jingjingwang: I think a better name might be `SequenceExpression`, as you already said in the javadoc :) "Counter" makes me think of counting something generated from its child, which is not what this expression does.

src/edu/washington/escience/myria/expression/BinaryExpression.java
```diff
@@ -223,25 +223,40 @@ protected final Type checkAndReturnDefaultNumericType(
...
-   * boolean.
+   * A function that could be used as the default type checker for a binary expression where each operand must be of a
+   * specified type.
    *
+   * @param leftExpectedType expected type of left operand
+   * @param rightExpectedType expected type of right operand
    * @param parameters parameters that are needed to determine the output type
    */
-  protected final void checkBooleanType(final ExpressionOperatorParameter parameters) {
+  protected final void checkOperandTypes(
```
jingjingwang: This is good but we can have something similar in `UnaryExpression`.

src/edu/washington/escience/myria/expression/CounterExpression.java
```diff
@@ -0,0 +1,76 @@
...
+   *
+   * @param operand an expression that evaluates to a positive integer
+   */
+  public CounterExpression(final ExpressionOperator operand) {
+    super(operand);
+  }
+
+  @Override
+  public Type getOutputType(final ExpressionOperatorParameter parameters) {
+    Type operandType = getOperand().getOutputType(parameters);
```
jingjingwang: We can add a `checkOperandType` in `UnaryExpression` and call it as we did for `BinaryExpression` and `SplitExpression`.

src/edu/washington/escience/myria/expression/ExpressionOperator.java
```diff
@@ -57,6 +57,7 @@
   @Type(name = "OR", value = OrExpression.class),
   @Type(name = "PLUS", value = PlusExpression.class),
   @Type(name = "POW", value = PowExpression.class),
+  @Type(name = "SPLIT", value = SplitExpression.class),
```
jingjingwang: Is `COUNTER` missing (in the UnaryExpression section above)?

src/edu/washington/escience/myria/expression/evaluate/GenericEvaluator.java
```diff
@@ -88,24 +96,26 @@ public void compile() throws DbException {
...
   /**
    * Evaluates the {@link #getJavaExpressionWithAppend()} using the {@link #evaluator}. Prefer to use
-   * {@link #evaluateColumn(TupleBatch)} as it can copy data without evaluating the expression.
+   * {@link #evaluateColumn(TupleBatch)} since it can evaluate an entire TupleBatch at a time for better locality.
    *
    * @param tb a tuple batch
-   * @param rowIdx the row that should be used for input data
-   * @param result the column that the result should be appended to
+   * @param rowIdx index of the row that should be used for input data
+   * @param count column storing number of results
```
jingjingwang: I'd like to add some check on `count == null` for single value expressions somewhere, or at least a comment on this prerequisite at a proper place. I'm not sure if here is the best place though.

src/edu/washington/escience/myria/expression/evaluate/GenericEvaluator.java
```diff
@@ -121,27 +131,93 @@ public String getJavaExpressionWithAppend() {
...
+
+    protected EvaluatorResult(
+        @Nonnull final ReadableColumn results, @Nonnull final Column<?> resultCounts) {
+      this.results = results;
+      this.resultCounts = resultCounts;
+    }
+    /**
+     * @return a {@link ReadableColumn} containing results from {@link #evaluateColumn}
+     */
+    public ReadableColumn getResults() {
```
jingjingwang: Seems this method is not needed because we can also pass `List<Column<?>>` to `outputBuffer.appendFromColumn()` in `Apply`.

src/edu/washington/escience/myria/expression/evaluate/GenericEvaluator.java
```diff
@@ -121,27 +131,93 @@ public String getJavaExpressionWithAppend() {
...
-
-    Type type = getOutputType();
-
-    ColumnBuilder<?> ret = ColumnFactory.allocateColumn(type);
-    for (int row = 0; row < tb.numTuples(); ++row) {
-      /** We already have an object, so we're not using the wrong version of put. Remove the warning. */
-      eval(tb, row, ret, null);
+    // For multivalued expressions, we may get more than `TupleBatch.BATCH_SIZE` results,
+    // so we need to allocate a `TupleBuffer` instead of a `ColumnBuilder` for results.
+    // (That is why we need `ReadableColumn` rather than `Column` as the result type.)
```
jingjingwang: By calling `TupleBuffer.finalResult()` we will get the list of finalized tuple batches which consist of `Column<?>`. Finalizing makes sense since `resultsBuffer` will not be changed after `eval()`s. Then we can avoid the awkwardness of having `Column` and `ReadableColumn` together.

src/edu/washington/escience/myria/expression/evaluate/GenericEvaluator.java
```diff
@@ -121,27 +131,93 @@ public String getJavaExpressionWithAppend() {
...
+      return results;
+    }
+    /**
+     * This is useful for, e.g., constructing a {@link TupleBatch} from several evaluators.
+     * This will return null if the result was not generated from a single-valued expression.
+     *
+     * @return a {@link Column<?>} containing results from {@link #evaluateColumn}
+     */
+    public Column<?> getResultsAsColumn() {
+      if (results instanceof Column<?>) {
```
jingjingwang: This is not needed anymore If we always return `Column`, see the comment in `GenericEvaluator.evaluateColumn()`.

src/edu/washington/escience/myria/operator/Apply.java
senderista: Yes there is, if we're generating the product in lexicographic order, which I want to do.

src/edu/washington/escience/myria/expression/evaluate/GenericEvaluator.java
senderista: So you're suggesting I call `getDataColumns()` on each `TupleBatch` from `finalResult()` and return this `List<Column>` in `EvaluatorResult.getResults()` (returning a singleton list for single-valued expressions)? That would definitely remove a lot of awkwardness.

src/edu/washington/escience/myria/expression/evaluate/GenericEvaluator.java
senderista: Not sure what you mean by that; we can only pass `ReadableColumn` to `appendFromColumn()`. Actually, I'm finding that returning `List<Column>` from `getResults()` is making things considerably more awkward in `Apply.fetchNextReady()`, because I now have to do my own row index management for `List<Column>` instead of letting `TupleBuffer` do it for me.

senderista: After taking your suggestion to return just `List<Column>` from this method, I ended up reverting to a similar interface as before. It's just too awkward to construct the Cartesian product with a `List<Column>` instead of a `TupleBuffer` (keeping in mind that these sequences of columns are "vertical", not "horizontal", since the `TupleBuffer` schema has only one field). The additional bookkeeping adds enough complexity to outweigh the simplification of the interface. Hopefully we can find a solution that simplifies both the interface definition and the client code...

src/edu/washington/escience/myria/operator/Apply.java
senderista: I understand that this seems inefficient, but it's only called once per `TupleBatch`, and I like to avoid materializing redundant state whenever possible. I don't have a very strong opinion, though.

src/edu/washington/escience/myria/expression/evaluate/GenericEvaluator.java
jingjingwang: I see what you meant. I agree with you that there is no easy way to avoid this without rethinking how to organize our current tuples/columns classes to provide better abstractions.

senderista:
I think I addressed all comments in 76c9ee4. I'm still not satisfied with the interface of `EvaluatorResult`, but maybe we can improve it after discussion in person.

jingjingwang:
Yes, maybe we need to do some refactoring to those column/tuple classes to help `EvaluatorResult`.

PR #737, opened 2015-04-28
do not call getSchema() in setChildren()

jingjingwang:
During a fragment instantiation, the order of calls to operators' connect() can be arbitrary. getSchema() goes to generateSchema() which may need to know the children of an operator that may not have been set by connect(). On the other hand, I think it is unnecessary to call getSchema() in setChildren() since we can always generate schema after childrens are being set.


jingjingwang:
@jortiz16 if you're interested in doing a little homework, I'd like to see your query being added as a test :) You can either submit the json query directly (example: JsonQuerySubmitTest) or manually specify such a plan (example: OperatorTestUsingSQLiteStorage), and check if it fails without the above commit.

PR #739, opened 2015-05-01
fix overflow bug with cosmo8

jortiz16:
When checking for the number of particles for star, dark and gas in the tipsy binary files, cosmo50 simulations would work just fine. Newer simulations (cosmo8) caused overflow issues during this check. Minor change was to cast the int multiplication/addition check to a long instead.


jortiz16:
Thanks @jingjingwang ! I changed the n* variables to longs instead. 

@domoritz I'm not familiar with that, but I think we talked about Guava during a previous Myria meeting? I don't remember what was said though

PR #740, opened 2015-05-05
Adding all *FileScan operators to list of singleton operators.

senderista:
I think omitting all the FileScan operators from this list was an oversight; at any rate this is causing duplicate output whenever a query with a FileScan source is run on a cluster with more than one worker. Not sure if there are any potential gotchas; @mbalazin mentioned that one of the astronomy formats (Tipsy?) might have files with the same name on different workers with different contents, so I'd like to confirm if that is current usage.

### DIFF
--- a/src/edu/washington/escience/myria/api/encoding/QueryConstruct.java
+++ b/src/edu/washington/escience/myria/api/encoding/QueryConstruct.java
@@ -317,7 +317,9 @@ public class QueryConstruct {
     for (PlanFragmentEncoding fragment : fragments) {
       for (OperatorEncoding<?> operator : fragment.operators) {
         if (operator instanceof CollectConsumerEncoding || operator instanceof SingletonEncoding
-            || operator instanceof EOSControllerEncoding) {
+            || operator instanceof EOSControllerEncoding || operator instanceof BinaryFileScanEncoding
+            || operator instanceof FileScanEncoding || operator instanceof NChiladaFileScanEncoding
+            || operator instanceof SeaFlowFileScanEncoding || operator instanceof TipsyFileScanEncoding) {
           if (fragment.workers == null) {
             fragment.workers = singletonWorkers;
           } else {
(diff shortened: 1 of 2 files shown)


BrandonHaynes:
I believe that I've implicitly relied on this behavior using relative import paths.

domoritz:
To be honest, I don't know this code at all and don't feel comfortable reviewing it. Also, the tests are failing.

senderista:
I just fixed the tests.

senderista:
@BrandonHaynes can we discuss this compatibility issue at the next Myria meeting?

domoritz:
@BrandonHaynes are you comfortable with reviewing this?

BrandonHaynes:
@senderista I'm not impacted by this change in any meaningful manner and don't feel that it warrants much delay.  (One can still build an equivalent plan with many FileScan operators).

@domoritz I've never visited QueryConstruct.java before, and am probably not the right person for this either.

senderista:
@jingjingwang please let me know if there are any blockers to merging this change this week.

jingjingwang:
I'd like to add such a comment somewhere saying we don't support loading from text files with the same name and path on different machines anymore since the semantics is not clear, QueryConstruct might be one place (and maybe a better error message for FileScan(s)), if you can think of other places that would be great too.

senderista:
Should this be a warning log message?

jingjingwang:
i guess it's an error not a warning since we'll reject the query if it happens..?

senderista:
Not sure what you mean, if the user assigns more than one worker ID in the
query plan then the existing code throws an exception, but that can't
happen if all query processing goes through RACO.

I just added a warning whenever we assign a random singleton worker to an
operator.

senderista:
@jingjingwang thanks!

@BrandonHaynes can we get this pushed to demo today so Magda can verify this works before next week? Do you need to fix anything on your side first that would be broken by this change?

PR #743, opened 2015-05-14
Delete tables

jortiz16:
fixes uwescience/myria-web#261


jortiz16:
Probably a silly question, but is there a way I can re-run these tests? (there was a timeout issue with the test that failed with the latest commit) When I run the systemtests from my side, everything passes. 

Thanks!

senderista:
Yes, I restarted the Travis build. There is a "replay" button at the upper
right corner of the Travis page, right below "Settings". (I couldn't find
it either until Dominik pointed it out to me.)

jortiz16:
Cool! Thanks!

src/edu/washington/escience/myria/accessmethod/AccessMethod.java
```diff
@@ -154,6 +154,12 @@ public void init() throws DbException {
   public abstract void dropTableIfExists(RelationKey relationKey) throws DbException;
 
   /**
+   * @param relationDelete the relation to drop
+   * @throws DbException if there is an error dropping the table.
+   */
+  public abstract void dropTableIfExistsCascade(RelationKey relationDelete) throws DbException;
```
senderista: Please rename the `relationDelete` param to `relationKey` for consistency with the other methods and the `JdbcAccessMethod` implementation.

src/edu/washington/escience/myria/accessmethod/JdbcAccessMethod.java
```diff
@@ -425,6 +425,21 @@ public void dropAndRenameTables(final RelationKey oldRelation, final RelationKey
   }
 
   @Override
+  public void dropTableIfExistsCascade(final RelationKey relationKey) throws DbException {
+    switch (jdbcInfo.getDbms()) {
+      case MyriaConstants.STORAGE_SYSTEM_MYSQL:
+        execute("DROP TABLE IF EXISTS " + quote(relationKey));
```
senderista: Should we log a warning that MySQL doesn't implement CASCADE?

src/edu/washington/escience/myria/accessmethod/SQLiteAccessMethod.java
```diff
@@ -455,6 +455,11 @@ public void createIndexIfNotExists(final RelationKey relationKey, final Schema s
       throws DbException {
     throw new UnsupportedOperationException("create index if not exists is not supported in sqlite yet, implement me");
   }
+
+  @Override
+  public void dropTableIfExistsCascade(final RelationKey relationDelete) throws DbException {
```
senderista: Please move this method adjacent to (immediately below) `dropTableIfExists` and rename `relationDelete` to `relationKey` per above.

src/edu/washington/escience/myria/parallel/Server.java
```diff
@@ -1033,6 +1034,45 @@ public void importDataset(final RelationKey relationKey, final Schema schema, fi
...
+
+    /* delete from postgres at each worker by calling the DbDelete operator */
+    try {
+      Map<Integer, SubQueryPlan> workerPlans = new HashMap<>();
+      for (Integer workerId : getWorkersForRelation(relationKey, null)) {
+        workerPlans.put(workerId, new SubQueryPlan(new DbDelete(EmptyRelation.of(catalog.getSchema(relationKey)),
+            relationKey, null)));
+      }
+      ListenableFuture<Query> qf =
+          queryManager.submitQuery("delete " + relationKey.toString(), "delete " + relationKey.toString(),
```
senderista: So I'm concerned about the transactional integrity of this distributed delete. I know we don't really do distributed transactions in Myria, but I don't really see a way around it here. I know the Postgres JDBC driver supports "XA" (two-phase commit), but I've never used it, and I'm not sure how it would map to our subquery/operator model. I know that `DROP TABLE` is transactional in Postgres, so at least you don't have to do `TRUNCATE TABLE` and `DROP TABLE` in separate steps.

src/edu/washington/escience/myria/operator/DbDelete.java
```diff
@@ -0,0 +1,64 @@
...
+import edu.washington.escience.myria.DbException;
+import edu.washington.escience.myria.MyriaConstants;
+import edu.washington.escience.myria.RelationKey;
+import edu.washington.escience.myria.accessmethod.AccessMethod;
+import edu.washington.escience.myria.accessmethod.ConnectionInfo;
+import edu.washington.escience.myria.storage.TupleBatch;
+
+/**
+ * 
+ * TODO: it's not really an "operator", is it?
```
senderista: Certainly making `DELETE` an "operator" is an awkward fit, but more importantly Myria's distributed query plans aren't transactional--see below.

src/edu/washington/escience/myria/coordinator/catalog/MasterCatalog.java
```diff
@@ -1753,6 +1753,40 @@ private void deleteRelationIfExists(@Nonnull final SQLiteConnection sqliteConnec
   }
 
   /**
+   * Connects to SQLite to delete
+   * 
+   * @throws CatalogException if there is an error in the catalog.
+   */
+  public void deleteRelationFromCatalog(final RelationKey relationkey) throws CatalogException {
```
senderista: `relationKey`

src/edu/washington/escience/myria/coordinator/catalog/MasterCatalog.java
```diff
@@ -1753,6 +1753,40 @@ private void deleteRelationIfExists(@Nonnull final SQLiteConnection sqliteConnec
...
+              sqliteConnection.exec("ROLLBACK TRANSACTION;");
+            } catch (SQLiteException e2) {
+              assert true; /* Do nothing */
+            }
+            throw e;
+          }
+          return null;
+        }
+      }).get();
+    } catch (InterruptedException | ExecutionException e) {
```
senderista: I don't think it ever makes sense to wrap an `InterruptedException`. Either put it in your `throws` clause (which you can't do if you're implementing `Runnable` but you can do with `Callable`), or call `Thread.currentThread().interrupt()` and swallow the exception. Either approach ensures that the caller (possibly scheduling code which you don't control) is properly notified of the interruption and can handle it appropriately (e.g., cancelling a `Future`). For example, perhaps someone wants to call *your* code in a `Future`--if you wrap `InterruptedException` they would see an interruption as an `ExecutionException` rather than an `InterruptedException`.

senderista: Actually, now that I think about it, only the visibility of the deleted relation needs to be atomic, and that can be accomplished just with a transactional update of the master catalog. The data deletion on the workers doesn't need to be transactional at all, and could simply retry until it succeeds. In fact, it could be asynchronous: the workers could periodically compare the local Postgres catalog to the master catalog and delete any tables corresponding to relations not present in the master catalog. Of course, we'd need to anticipate possible race conditions--perhaps a Postgres advisory lock that would need to be acquired by workers to either create local partitions or delete them would work. These are just some half-baked ideas--I guess we should discuss the design in the next Myria meeting.

src/edu/washington/escience/myria/parallel/Server.java
jortiz16: I've never used the two-phase commit either, but I think this comment ties along with what you mentioned at the comment under ```MasterCatalog.java``` (the interaction between the master catalog and what relations the workers currently have -- at creation and deletion).

src/edu/washington/escience/myria/accessmethod/JdbcAccessMethod.java
jortiz16: Sure, I can add it. I'm not even sure people use MySQL as the underlying db for Myria anymore? I thought most people used SQLite and Postgres for the most part.  I just added a MySQL condition because the other methods in ```JdbcAccessMethod.java``` referred to it too.

src/edu/washington/escience/myria/coordinator/catalog/MasterCatalog.java
jortiz16: Sure, I'd be happy to talk about it in the meeting too. If we end up doing asynchronous checking and comparing between local and master catalogs, then wouldn't the methods that create relations also need to be changed as well? I like the idea of locking though too. Again, we can bring this up at the meeting.

jortiz16:
Sorry for the delay. I made the change with the exception (I hope that is what you meant) and I also added some tests. 

I talked to Magda about your suggestions with potentially adding a 2PC protocol in order to make sure that relations get deleted correctly. She mentioned that for now, we should only make sure that the call is idempotent. So now, as long as the relation exists in the catalog, we should be able to call this method even if not all the workers have the underlying relation we are trying to delete. I tried to add such an example in the tests. What do you think?

jortiz16:
tests are failing because I am using an old version of SystemTestBase it seems. I'll fix this once I get to campus.

jortiz16:
I had to run a rebase to get the new tests to work. I didn't know it would show all my commits again on here. Relevant commits start from 7da559d17425dbcab799bae7cbf628b0033de60d onwards.

jortiz16:
I added the tombstone column along with the queries from the MasterCatalog that were called from the DatasetResource class. Once a relation is tombstoned, the goal is to not have it show up through the API calls from DatasetResource (i.e. getDatasets, getDatasetsForUsers, etc).

src/edu/washington/escience/myria/accessmethod/JdbcAccessMethod.java
```diff
@@ -425,6 +425,22 @@ public void dropAndRenameTables(final RelationKey oldRelation, final RelationKey
...
+    switch (jdbcInfo.getDbms()) {
+      case MyriaConstants.STORAGE_SYSTEM_MYSQL:
+        LOGGER.warn("MySQL does not implement CASCADE");
+        execute("DROP TABLE IF EXISTS " + quote(relationKey));
+        break;
+      case MyriaConstants.STORAGE_SYSTEM_POSTGRESQL:
+        execute("DROP TABLE IF EXISTS " + quote(relationKey) + " CASCADE");
+        break;
+      default:
+        throw new UnsupportedOperationException("Don't know whether " + jdbcInfo.getDbms()
```
senderista: Add `CASCADE` to the message.

src/edu/washington/escience/myria/accessmethod/JdbcAccessMethod.java
```diff
@@ -425,6 +425,22 @@ public void dropAndRenameTables(final RelationKey oldRelation, final RelationKey
   }
 
   @Override
+  public void dropTableIfExistsCascade(final RelationKey relationKey) throws DbException {
+    switch (jdbcInfo.getDbms()) {
+      case MyriaConstants.STORAGE_SYSTEM_MYSQL:
+        LOGGER.warn("MySQL does not implement CASCADE");
```
senderista: I'm not sure that just logging a warning is the right approach since it might never be seen. I think if we're violating the user's expectations then the correct thing to do is throw an exception. This may be all moot for now given that we don't use MySQL, but who knows what external users might want to do...

src/edu/washington/escience/myria/accessmethod/SQLiteAccessMethod.java
```diff
@@ -401,6 +401,11 @@ public void dropTableIfExists(final RelationKey relationKey) throws DbException
   }
 
   @Override
+  public void dropTableIfExistsCascade(final RelationKey relationKey) throws DbException {
```
senderista: Shouldn't you at least log a warning that SQLite doesn't support `CASCADE`?

src/edu/washington/escience/myria/coordinator/catalog/MasterCatalog.java
```diff
@@ -117,6 +115,7 @@
     + "    program_name TEXT NOT NULL,\n"
     + "    relation_name TEXT NOT NULL,\n"
     + "    num_tuples INTEGER NOT NULL,\n"
+    + "    tombstone INTEGER NOT NULL,\n"
```
senderista: I think a less jargon-ish and more descriptive name than `tombstone` would be appropriate, maybe `is_deleted` or `deleting` or `marked_for_deletion`?

src/edu/washington/escience/myria/coordinator/catalog/MasterCatalog.java
```diff
@@ -1790,6 +1790,9 @@ protected Long job(final SQLiteConnection sqliteConnection) throws CatalogExcept
         }
       }).get();
     } catch (InterruptedException | ExecutionException e) {
+      if (e instanceof InterruptedException) {
```
senderista: Sorry, I was completely wrong about my suggested handling of `InterruptedException` (@jingjingwang straightened me out). It doesn't make sense here to set the interrupted status of the current thread, because the `InterruptedException` was caught on a completely different thread (the thread pool worker where the `Future` was executed). So the correct thing to do here is just what the original code was doing.

src/edu/washington/escience/myria/coordinator/catalog/MasterCatalog.java
```diff
@@ -889,7 +889,7 @@ private Schema getSchema(@Nonnull final SQLiteConnection sqliteConnection, @Nonn
     try {
       SQLiteStatement statement =
           sqliteConnection
-              .prepare("SELECT col_name,col_type FROM relation_schema WHERE user_name=? AND program_name=? AND relation_name=?; ORDER BY col_index ASC");
+              .prepare("SELECT col_name,col_type FROM relation_schema JOIN relations USING (user_name,program_name,relation_name) WHERE user_name=? AND program_name=? AND relation_name=? ORDER BY col_index ASC");
```
senderista: Why are you joining to the `relations` table but not filtering on the `tombstone` attribute?

jortiz16: My fault. I initially had the ```tombstone``` filter but I took it out because my delete call calls getSchema() at some point.  I forgot to take out the ``JOIN relations```

src/edu/washington/escience/myria/coordinator/catalog/MasterCatalog.java
jortiz16: Okay will do!

src/edu/washington/escience/myria/coordinator/catalog/MasterCatalog.java
jortiz16: No problem, I'll rename to ```is_deleted```

src/edu/washington/escience/myria/coordinator/catalog/MasterCatalog.java
```diff
@@ -1805,7 +1808,8 @@ private void deleteRelationIfExists(@Nonnull final SQLiteConnection sqliteConnec
       @Nonnull final RelationKey relation) throws CatalogException {
     try {
       SQLiteStatement statement =
-          sqliteConnection.prepare("DELETE FROM relations WHERE user_name=? AND program_name=? AND relation_name=?;");
+          sqliteConnection
+              .prepare("DELETE FROM relations WHERE user_name=? AND program_name=? AND relation_name=? AND tombstone=1;");
```
senderista: Where do we delete the corresponding entries from the `relation_schema` table?

src/edu/washington/escience/myria/coordinator/catalog/MasterCatalog.java
```diff
@@ -1818,6 +1822,79 @@ private void deleteRelationIfExists(@Nonnull final SQLiteConnection sqliteConnec
...
+        protected Void job(final SQLiteConnection sqliteConnection) throws CatalogException, SQLiteException {
+          sqliteConnection.exec("BEGIN TRANSACTION;");
+          try {
+            deleteRelationIfExists(sqliteConnection, relationKey);
+            sqliteConnection.exec("COMMIT TRANSACTION;");
+          } catch (SQLiteException e) {
+            try {
+              sqliteConnection.exec("ROLLBACK TRANSACTION;");
+            } catch (SQLiteException e2) {
+              assert true; /* Do nothing */
```
senderista: Shouldn't we log a warning at least?

src/edu/washington/escience/myria/coordinator/catalog/MasterCatalog.java
```diff
@@ -1891,7 +1968,7 @@ protected Void job(final SQLiteConnection sqliteConnection) throws CatalogExcept
           try {
             SQLiteStatement statement =
                 sqliteConnection
-                    .prepare("UPDATE relations SET num_tuples=? WHERE user_name=? AND program_name=? AND relation_name=?;");
+                    .prepare("UPDATE relations SET num_tuples=? WHERE user_name=? AND program_name=? AND relation_name=? and tombstone=0;");
```
senderista: Why would we want to silently fail here? Why not throw an exception on any update operations to tombstoned relation metadata?

src/edu/washington/escience/myria/coordinator/catalog/MasterCatalog.java
jortiz16: On MasterCatalog, the ```relation_schema``` table is declared with the following constraint.... ```FOREIGN KEY (user_name,program_name,relation_name) REFERENCES relations ON DELETE CASCADE);```

So if ``relations``` deletes a row, it should be deleted from ```relation_schema``` and  ```stored_relations``` as well I believe

src/edu/washington/escience/myria/accessmethod/JdbcAccessMethod.java
jortiz16: Yeah, I don't think our group uses MySQL, but it doesn't hurt to add a warning here

jortiz16: Okay I just looked at it again (I wasn't sure earlier). For now, I'll just add an ```UnsupportedOperationException```  for this case

src/edu/washington/escience/myria/coordinator/catalog/MasterCatalog.java
jortiz16: A warning about which part? If the transaction rolls back?

src/edu/washington/escience/myria/coordinator/catalog/MasterCatalog.java
jortiz16: How could we handle this? Should there be a query before this one which checks if the RelationKey is already tombstoned? If so, then return a ```CatalogException```?

senderista: Yes, that's what I was thinking.

src/edu/washington/escience/myria/operator/DbDelete.java
```diff
@@ -0,0 +1,74 @@
...
+  protected void init(final ImmutableMap<String, Object> execEnvVars) throws DbException {
+    /* Retrieve connection information from the environment variables, if not already set */
+    if (connectionInfo == null && execEnvVars != null) {
+      connectionInfo = (ConnectionInfo) execEnvVars.get(MyriaConstants.EXEC_ENV_VAR_DATABASE_CONN_INFO);
+    }
+    /* Open the database connection */
+    accessMethod = AccessMethod.of(connectionInfo.getDbms(), connectionInfo, false);
+
+    /* Drop the table */
+    accessMethod.dropTableIfExistsCascade(relationKey);
```
senderista: Won't we throw an exception here if the relation doesn't exist on this worker? What about cases where the relation was only partially deleted (i.e., some partitions still exist)--don't we want the subqueries on workers where the partition was already deleted to succeed, to ensure idempotence?

src/edu/washington/escience/myria/parallel/Server.java
```diff
@@ -1036,6 +1037,52 @@ public void importDataset(final RelationKey relationKey, final Schema schema, fi
...
+              "deleting from " + relationKey.toString(getDBMS()), new SubQueryPlan(new SinkRoot(new EOSSource())),
+              workerPlans);
+      try {
+        qf.get();
+      } catch (ExecutionException e) {
+        throw new DbException("Error executing query", e.getCause());
```
senderista: If the DbDelete operator fails when the partition doesn't exist on a worker, don't we need to handle the relevant exception here to ensure the API call succeeds when it is retried after a partial failure (i.e., some but not all partitions were already deleted)?

src/edu/washington/escience/myria/util/SQLiteUtils.java
```diff
@@ -119,4 +119,81 @@ public static void createTable(final String dbFileAbsolutePath, final RelationKe
...
+
+      if (!f.getParentFile().exists()) {
+        f.getParentFile().mkdirs();
+      }
+
+      /* Connect to the database */
+      sqliteConnection = new SQLiteConnection(f);
+      sqliteConnection.open(true);
+      statement =
+          sqliteConnection.prepare("SELECT * FROM sqlite_master WHERE type='table' AND name="
```
senderista: Don't you need to quote the table name as a string literal, not a quoted identifier?

src/edu/washington/escience/myria/util/SQLiteUtils.java
```diff
@@ -119,4 +119,81 @@ public static void createTable(final String dbFileAbsolutePath, final RelationKe
...
+   * @param relationKey the relation key to delete
+   * */
+  public static void deleteTable(final String dbFileAbsolutePath, final RelationKey relationKey) throws IOException,
+      SQLiteException {
+    SQLiteConnection sqliteConnection = null;
+    SQLiteStatement statement = null;
+    try {
+      final File f = new File(dbFileAbsolutePath);
+
+      if (!f.getParentFile().exists()) {
```
senderista: How could this ever be a valid failure scenario? Aren't we assuming that the database file exists?

src/edu/washington/escience/myria/util/SQLiteUtils.java
```diff
@@ -119,4 +119,81 @@ public static void createTable(final String dbFileAbsolutePath, final RelationKe
...
+   * */
+  public static boolean existsTable(final String dbFileAbsolutePath, final RelationKey relationKey) throws IOException,
+      SQLiteException {
+    boolean exists = false;
+    SQLiteConnection sqliteConnection = null;
+    SQLiteStatement statement = null;
+    try {
+      final File f = new File(dbFileAbsolutePath);
+
+      if (!f.getParentFile().exists()) {
```
senderista: How could this ever be a valid failure scenario? Shouldn't any caller be able to assume that the database file exists?

src/edu/washington/escience/myria/util/SQLiteUtils.java
jortiz16: Yes, but I included this method to check if the delete call still works even though some of the workers do not have that relation. See DbDeleteTest.java (third scenario)

systemtest/edu/washington/escience/myria/systemtest/DbDeleteTest.java
```diff
@@ -0,0 +1,124 @@
...
+  public void testDeleteRelationInCatalog() throws Exception {
+    ingestTestDataset();
+
+    Set<Integer> workers = server.getWorkersForRelation(relationKey, null);
+    assertTrue(workers.size() == workerIDs.length);
+
+    JsonAPIUtils.deleteDataset("localhost", masterDaemonPort, relationKey.getUserName(), relationKey.getProgramName(),
+        relationKey.getRelationName());
+
+    workers = server.getWorkersForRelation(relationKey, null);
```
senderista: Does this actually test if the relation has been deleted from the catalog entirely, or just that no partitions exist?

systemtest/edu/washington/escience/myria/systemtest/DbDeleteTest.java
```diff
@@ -0,0 +1,124 @@
...
+    assertFalse(existsTable(workerIDs[1], relationKey));
+  }
+
+  /**
+   * Ingest a test dataset.
+   */
+  public void ingestTestDataset() throws Exception {
+    DataSource relationSource = new FileSource(Paths.get("testdata", "filescan", "simple_two_col_int.txt").toString());
+    relationKey = RelationKey.of("public", "adhoc", "testIngest");
+    relationSchema = Schema.of(ImmutableList.of(Type.INT_TYPE, Type.INT_TYPE), ImmutableList.of("x", "y"));
```
senderista: `Schema.ofFields()`

src/edu/washington/escience/myria/util/SQLiteUtils.java
jortiz16: This method and the ```deleteTables``` method are not failure scenarios. I used them as helper methods for DbDeleteTest.java

src/edu/washington/escience/myria/coordinator/catalog/MasterCatalog.java
senderista: No, if an exception is thrown during rollback.

systemtest/edu/washington/escience/myria/systemtest/DbDeleteTest.java
jortiz16: Okay, I'll make this more explicit

jortiz16:
Hey Tobin, I added back ```cascade delete```, but I wasn't sure what you meant by adding a "best effort" parameter?

senderista:
Just meant a boolean parameter to simply log a warning if cascade isn't supported rather than throw an exception.

domoritz:
@senderista Awesome. Can you ping myria@ when this is deployed so that we can all start deleting tables?

PR #745, opened 2015-05-15
Setting bootclasspath for cross-compiling on Java 8.

senderista:
After compiling Myria binaries on my laptop with Java 8 installed, I got mysterious errors after deployment which turned out to be due to not setting the `-bootclasspath` option to` javac`. See this article for explanation: https://blogs.oracle.com/darcy/entry/how_to_cross_compile_for. To cross-compile correctly from a newer JDK version to an older one, it is not enough to set `-source` and `-target` options as we do currently; we also have to set the boot classpath (where the `javac` class loader looks for "bootstrap" libraries). With this change, to build Myria on a system with Java 8 installed, you must install JDK 7 and set the `JAVA7_HOME` environment variable to the location of your JDK. On a Mac, this can be done properly by adding the following to your `.zshenv` or `.bashrc`:
`export JAVA7_HOME=$(/usr/libexec/java_home -v 1.7)`

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -170,6 +170,18 @@ dependencies {
   testsuitesRuntime configurations.systemtestRuntime
 }
 
+/* If Java 8 is the default JDK but Java 7 is installed, do a proper cross-compile. */
+tasks.withType(JavaCompile) {
+  doFirst {
+    if (sourceCompatibility == '1.7' && System.env.JAVA7_HOME != null) {
+      options.fork = true
+      options.bootClasspath = "$System.env.JAVA7_HOME/jre/lib/rt.jar"
+      options.bootClasspath += "$File.pathSeparator$System.env.JAVA7_HOME/jre/lib/jsse.jar"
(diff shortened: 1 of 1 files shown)


jingjingwang:
Also, from the link that you gave, seems not doing a cross-compile is the best way: 
"The most reliably way to produce class files that will work on a particular JDK and later is to compile the source files using the oldest JDK of interest. ..."

senderista:
I've tried the other approach of setting JAVA_HOME to the non-default JDK for a particular build and it was far more trouble than this cross-compiling approach.

senderista:
Also, in `gradle.properties` we could set `org.gradle.java.home=$JAVA7_HOME`, but this still requires the user to set an environment variable for the location of their alternative JDK, so it's no less work. On the other hand, it might be more robust than specifying the boot classpath and using a newer JDK for the build--I'm not really sure. I'll try it locally and see if it works on a Java7 runtime environment (EC2).

PR #746, opened 2015-05-16
adding streaming aggregate & limit operator

yqguo:
adding streaming aggregate class and test, modified Limit.java to stop ingesting tuples from child when reached limit

### DIFF
--- a/src/edu/washington/escience/myria/operator/Limit.java
+++ b/src/edu/washington/escience/myria/operator/Limit.java
@@ -27,9 +27,6 @@ public final class Limit extends UnaryOperator {
   /** The number of tuples left to emit. */
   private long toEmit;
 
-  /** If number of emitted tuples reached limit. */
-  private boolean done;
-
   /**
    * A limit operator keeps the first <code>limit</code> tuples produced by its child.
    * 
@@ -41,30 +38,28 @@ public final class Limit extends UnaryOperator {
     this.limit = Objects.requireNonNull(limit, "limit");
     Preconditions.checkArgument(limit >= 0L, "limit must be non-negative");
     toEmit = this.limit;
-    done = false;
   }
 
   @Override
   protected TupleBatch fetchNextReady() throws DbException {
     Operator child = getChild();
-    if (done) {
-      return null;
-    } else {
-      TupleBatch tb = child.nextReady();
-      TupleBatch result = null;
-      if (tb != null) {
-        if (tb.numTuples() <= toEmit) {
-          toEmit -= tb.numTuples();
-          result = tb;
-        } else if (toEmit > 0) {
-          result = tb.prefix(Ints.checkedCast(toEmit));
-          toEmit = 0;
-          child.close();
-          done = true;
-        }
+    TupleBatch tb = child.nextReady();
+    TupleBatch result = null;
+    if (tb != null) {
+      if (tb.numTuples() <= toEmit) {
+        toEmit -= tb.numTuples();
+        result = tb;
+      } else if (toEmit > 0) {
+        result = tb.prefix(Ints.checkedCast(toEmit));
+        toEmit = 0;
       }
-      return result;
+      if (toEmit == 0) {
+        /* Close child and self. No more stream is needed. */
+        child.close();
+        close();
+      }
     }
+    return result;
   }
 
   @Override
--- a/src/edu/washington/escience/myria/operator/Limit.java
+++ b/src/edu/washington/escience/myria/operator/Limit.java
@@ -43,23 +43,25 @@ public final class Limit extends UnaryOperator {
   @Override
   protected TupleBatch fetchNextReady() throws DbException {
     Operator child = getChild();
-    TupleBatch tb = child.nextReady();
-    TupleBatch result = null;
-    if (tb != null) {
-      if (tb.numTuples() <= toEmit) {
-        toEmit -= tb.numTuples();
-        result = tb;
-      } else if (toEmit > 0) {
-        result = tb.prefix(Ints.checkedCast(toEmit));
-        toEmit = 0;
+    if (child.isOpen()) {
+      TupleBatch tb = child.nextReady();
+      TupleBatch result = null;
+      if (tb != null) {
+        if (tb.numTuples() <= toEmit) {
+          toEmit -= tb.numTuples();
+          result = tb;
+        } else if (toEmit > 0) {
+          result = tb.prefix(Ints.checkedCast(toEmit));
+          toEmit = 0;
+        }
+        if (toEmit == 0) {
+          /* Close child. No more stream is needed. */
+          child.close();
+        }
       }
-      if (toEmit == 0) {
-        /* Close child and self. No more stream is needed. */
-        child.close();
-        close();
-      }
+      return result;
     }
-    return result;
+    return null;
   }
 
   @Override
--- a/src/edu/washington/escience/myria/operator/Limit.java
+++ b/src/edu/washington/escience/myria/operator/Limit.java
@@ -12,9 +12,9 @@ import edu.washington.escience.myria.Schema;
 import edu.washington.escience.myria.storage.TupleBatch;
 
 /**
- * A poor implementation of a Limit operator, which emits the first N tuples then drops the rest on the floor.
+ * A poor implementation of a Limit operator, which emits the first N tuples then closes the child operator from further
+ * feeding tuples.
  * 
- * We would prefer one that stops the incoming stream, but this is not currently supported.
  */
 public final class Limit extends UnaryOperator {
 
@@ -27,6 +27,9 @@ public final class Limit extends UnaryOperator {
   /** The number of tuples left to emit. */
   private long toEmit;
 
+  /** If number of emitted tuples reached limit. */
+  private boolean done;
+
   /**
    * A limit operator keeps the first <code>limit</code> tuples produced by its child.
    * 
@@ -38,23 +41,30 @@ public final class Limit extends UnaryOperator {
     this.limit = Objects.requireNonNull(limit, "limit");
     Preconditions.checkArgument(limit >= 0L, "limit must be non-negative");
     toEmit = this.limit;
+    done = false;
   }
 
   @Override
   protected TupleBatch fetchNextReady() throws DbException {
     Operator child = getChild();
-    for (TupleBatch tb = child.nextReady(); tb != null; tb = child.nextReady()) {
-      if (tb.numTuples() <= toEmit) {
-        toEmit -= tb.numTuples();
-        return tb;
-      } else if (toEmit > 0) {
-        tb = tb.prefix(Ints.checkedCast(toEmit));
-        toEmit = 0;
-        return tb;
+    if (done) {
+      return null;
+    } else {
+      TupleBatch tb = child.nextReady();
+      TupleBatch result = null;
+      if (tb != null) {
+        if (tb.numTuples() <= toEmit) {
+          toEmit -= tb.numTuples();
+          result = tb;
+        } else if (toEmit > 0) {
+          result = tb.prefix(Ints.checkedCast(toEmit));
+          toEmit = 0;
+          child.close();
+          done = true;
+        }
       }
-      /* Else, drop on the floor. */
+      return result;
     }
-    return null;
   }
 
   @Override
--- a/src/edu/washington/escience/myria/operator/agg/StreamingAggregate.java
+++ b/src/edu/washington/escience/myria/operator/agg/StreamingAggregate.java
@@ -89,7 +89,7 @@ public class StreamingAggregate extends UnaryOperator {
   @Override
   protected TupleBatch fetchNextReady() throws DbException {
     if (child.eos()) {
-      return resultBuffer.popAny();
+      return null;
     }
     if (tb == null) {
       tb = child.nextReady();
@@ -170,7 +170,7 @@ public class StreamingAggregate extends UnaryOperator {
     }
 
     groupSchema = inputSchema.getSubSchema(gFields);
-    resultSchema = groupSchema;
+    resultSchema = Schema.of(groupSchema.getColumnTypes(), groupSchema.getColumnNames());
     try {
       for (Aggregator agg : AggUtils.allocateAggs(factories, inputSchema)) {
         Schema curAggSchema = agg.getResultSchema();
--- a/src/edu/washington/escience/myria/operator/agg/StreamingAggregate.java
+++ b/src/edu/washington/escience/myria/operator/agg/StreamingAggregate.java
@@ -1,0 +1,277 @@
+package edu.washington.escience.myria.operator.agg;
+
+import java.util.Objects;
+
+import javax.annotation.Nullable;
+
+import com.google.common.base.Preconditions;
+import com.google.common.collect.ImmutableList;
+import com.google.common.collect.ImmutableMap;
+
+import edu.washington.escience.myria.DbException;
+import edu.washington.escience.myria.Schema;
+import edu.washington.escience.myria.Type;
+import edu.washington.escience.myria.operator.Operator;
+import edu.washington.escience.myria.operator.UnaryOperator;
+import edu.washington.escience.myria.storage.Tuple;
+import edu.washington.escience.myria.storage.TupleBatch;
+import edu.washington.escience.myria.storage.TupleBuffer;
+import edu.washington.escience.myria.storage.TupleUtils;
+
+/**
+ * This aggregate operator computes the aggregation in streaming manner (requires input sorted on grouping column(s)).
+ * This supports aggregation over multiple columns, with one or more group by columns. Intend to substitute
+ * SingleGroupByAggregate and MultiGroupByAggregate when input is known to be sorted.
+ *
+ * @see Aggregate
+ * @see SingleGroupByAggregate
+ * @see MultiGroupByAggregate
+ */
+public class StreamingAggregate extends UnaryOperator {
+
+  /** Required for Java serialization. */
+  private static final long serialVersionUID = 1L;
+
+  /** The schema of the aggregation result. */
+  private Schema aggSchema;
+  /** The schema of the columns indicated by the group keys. */
+  private Schema groupSchema;
+  /** Holds the current grouping key. */
+  private Tuple curGroupKey;
+
+  /** Group fields. **/
+  private final int[] gFields;
+  /** Group field types. **/
+  private final Type[] gTypes;
+  /** An array [0, 1, .., gFields.length-1] used for comparing tuples. */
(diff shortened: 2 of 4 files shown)


src/edu/washington/escience/myria/operator/Limit.java
```diff
@@ -38,23 +41,30 @@ public Limit(@Nonnull final Long limit, final Operator child) {
     this.limit = Objects.requireNonNull(limit, "limit");
     Preconditions.checkArgument(limit >= 0L, "limit must be non-negative");
     toEmit = this.limit;
+    done = false;
```
senderista: I don't think you need this new flag. You should immediately return `null` from `fetchNextReady()` if the child has already been closed, which you can test with `child.isOpen()` or `child.eos()`. @jingjingwang do you agree?

src/edu/washington/escience/myria/operator/Limit.java
```diff
@@ -38,23 +41,30 @@ public Limit(@Nonnull final Long limit, final Operator child) {
...
     Operator child = getChild();
-    for (TupleBatch tb = child.nextReady(); tb != null; tb = child.nextReady()) {
-      if (tb.numTuples() <= toEmit) {
-        toEmit -= tb.numTuples();
-        return tb;
-      } else if (toEmit > 0) {
-        tb = tb.prefix(Ints.checkedCast(toEmit));
-        toEmit = 0;
-        return tb;
+    if (done) {
```
senderista: Replace this check with `child.isOpen()` or `child.eos()`. Actually, I think that this check is strictly unnecessary since after we've closed the child, its EOS status will be propagated to this operator and `nextReady()` will return `null` immediately. @jingjingwang could you comment?

src/edu/washington/escience/myria/operator/Limit.java
```diff
@@ -38,23 +41,30 @@ public Limit(@Nonnull final Long limit, final Operator child) {
...
+      return null;
+    } else {
+      TupleBatch tb = child.nextReady();
+      TupleBatch result = null;
+      if (tb != null) {
+        if (tb.numTuples() <= toEmit) {
+          toEmit -= tb.numTuples();
+          result = tb;
+        } else if (toEmit > 0) {
```
senderista: This logic is broken if `toEmit == 0` (implying that `tb.numTuples() == toEmit` on the previous call). In that case, you will return `null` on this and all subsequent calls until the child is exhausted, just like the previous `Limit` implementation. You need to close the child if `toEmit == 0` and return `null`. Also, please write a unit test which fails under this condition with the current implementation (just use a `TupleSource` containing as many tuples as the limit).

senderista: Actually I see that there's already a unit test as I described: `testLimitNumTuples`. It won't expose the bug though, since it doesn't test if the child has been closed after the limit is reached but before it is exhausted. You'll need something a little trickier to expose this bug--I would recommend using the `TupleSource(List<TupleBatch>)` constructor. You can generate two small `TupleBatch`es by creating two `TupleBatchBuffer`s with `TestUtils.range(limit)` and calling `popAny()` on each. Then initialize a `TupleSource` with a `List` of these two `TupleBatch`es and initialize a `Limit` operator with this `TupleSource`. Now call `nextReady()`, verify the `TupleBatch` returned has `numTuples() == limit`, and verify that the operator is at end-of-stream (`eos()`), which will be the case if it closed its child after the first call to `nextReady()`, as it should have.

@jingjingwang do you have a better idea for this test?

test/edu/washington/escience/myria/operator/StreamingAggTest.java
```diff
@@ -0,0 +1,1113 @@
...
+
+  /**
+   * Construct a TupleBatchBuffer to be used as source of aggregate. Fixed schema and sorted on grouping columns.
+   *
+   * @param numTuples number of tuples to be added
+   * @return filled TupleBatchBuffer with each group key having (numTuples/10) tuples
+   */
+  private TupleBatchBuffer fillInputTbb(final int numTuples) {
+    final Schema schema =
+        new Schema(ImmutableList.of(Type.INT_TYPE, Type.DOUBLE_TYPE, Type.FLOAT_TYPE, Type.LONG_TYPE,
```
senderista: Readability might be improved by using the `Schema.ofFields(name1, type1, name2, type2, ...)` factory method.

src/edu/washington/escience/myria/operator/Limit.java
yqguo: I see. I'll fix that and the tests. Thanks!

src/edu/washington/escience/myria/operator/agg/StreamingAggregate.java
```diff
@@ -0,0 +1,277 @@
...
+   * @param child The Operator that is feeding us tuples.
+   * @param gfields The columns over which we are grouping the result.
+   * @param factories The factories that will produce the {@link Aggregator}s for each group.
+   */
+  public StreamingAggregate(@Nullable final Operator child, final int[] gfields, final AggregatorFactory... factories) {
+    super(child);
+    gFields = Objects.requireNonNull(gfields, "gfields");
+    gTypes = new Type[gfields.length];
+    this.factories = Objects.requireNonNull(factories, "factories");
+    Preconditions.checkArgument(gfields.length > 0, " must have at least one group by field");
```
senderista: Pick either `length > 0` or `length != 0` and stick to it.

src/edu/washington/escience/myria/operator/agg/StreamingAggregate.java
```diff
@@ -0,0 +1,277 @@
...
+  /**
+   * Groups the input tuples according to the specified grouping fields, then produces the specified aggregates.
+   *
+   * @param child The Operator that is feeding us tuples.
+   * @param gfields The columns over which we are grouping the result.
+   * @param factories The factories that will produce the {@link Aggregator}s for each group.
+   */
+  public StreamingAggregate(@Nullable final Operator child, final int[] gfields, final AggregatorFactory... factories) {
+    super(child);
+    gFields = Objects.requireNonNull(gfields, "gfields");
```
senderista: You should also use the @Nonnull annotation on parameters that are required to be non-null, for the benefit of static analysis tools like FindBugs (which we run on the Myria codebase).

src/edu/washington/escience/myria/operator/agg/StreamingAggregate.java
```diff
@@ -0,0 +1,275 @@
...
+  public StreamingAggregate(@Nullable final Operator child, @Nonnull final int[] gfields,
+      @Nonnull final AggregatorFactory... factories) {
+    super(child);
+    gFields = Objects.requireNonNull(gfields, "gfields");
+    gTypes = new Type[gfields.length];
+    this.factories = Objects.requireNonNull(factories, "factories");
+    Preconditions.checkArgument(gfields.length > 0, " must have at least one group by field");
+    Preconditions.checkArgument(factories.length > 0, "to use StreamingAggregate, must specify some aggregates");
+    gRange = new int[gfields.length];
+    for (int i = 0; i < gfields.length; ++i) {
```
senderista: Nitpick: I'd prefer `gRange.length` here since the array bounds should be those of the array being modified.

src/edu/washington/escience/myria/operator/agg/StreamingAggregate.java
```diff
@@ -0,0 +1,275 @@
...
+    try {
+      for (Aggregator agg : AggUtils.allocateAggs(factories, inputSchema)) {
+        Schema curAggSchema = agg.getResultSchema();
+        aggTypes.addAll(curAggSchema.getColumnTypes());
+        aggNames.addAll(curAggSchema.getColumnNames());
+      }
+    } catch (DbException e) {
+      throw new RuntimeException("unable to allocate aggregators to determine output schema", e);
+    }
+    aggSchema = new Schema(aggTypes, aggNames);
```
senderista: You shouldn't be using the `Schema` constructor; instead use `Schema.of()`.

src/edu/washington/escience/myria/operator/agg/StreamingAggregate.java
```diff
@@ -0,0 +1,275 @@
...
+      return null;
+    }
+    Schema inputSchema = child.getSchema();
+    if (inputSchema == null) {
+      return null;
+    }
+
+    groupSchema = inputSchema.getSubSchema(gFields);
+
+    /* Build the output schema from the group schema and the aggregates. */
```
senderista: Why can't you simply construct `groupSchema` and then concatenate the schema for each `Aggregator` by calling `Schema.merge` in a loop?

src/edu/washington/escience/myria/operator/agg/StreamingAggregate.java
```diff
@@ -0,0 +1,275 @@
...
+  @Override
+  protected void init(final ImmutableMap<String, Object> execEnvVars) throws DbException {
+    Preconditions.checkState(getSchema() != null, "unable to determine schema in init");
+    aggregators = AggUtils.allocateAggs(factories, getChild().getSchema());
+    aggregatorStates = AggUtils.allocateAggStates(aggregators);
+    resultBuffer = new TupleBuffer(getSchema());
+  }
+
+  @Override
+  protected void cleanup() throws DbException {
```
senderista: What is the rationale for all this `null`-setting? I don't see how it's preventing any memory leaks...

src/edu/washington/escience/myria/operator/agg/StreamingAggregate.java
```diff
@@ -0,0 +1,275 @@
...
+    return null;
+  }
+
+  /**
+   * Re-initialize aggregator states for new group key.
+   *
+   * @throws DbException if any error
+   */
+  private void reinitializeAggStates() throws DbException {
+    aggregatorStates = null;
```
senderista: not necessary

src/edu/washington/escience/myria/operator/agg/StreamingAggregate.java
```diff
@@ -0,0 +1,275 @@
...
+   * Returns the next tuple batch containing the result of this aggregate. Grouping field(s) followed by aggregate
+   * field(s).
+   *
+   * @throws DbException if any error occurs.
+   * @return result tuple batch
+   */
+  @Override
+  protected TupleBatch fetchNextReady() throws DbException {
+    final Operator child = getChild();
+    if (child.eos()) {
```
senderista: Why can't you always return the last TupleBatch you finalized? Isn't that what "streaming" means (i.e., not waiting to return results until the child is exhausted)?

src/edu/washington/escience/myria/operator/agg/StreamingAggregate.java
```diff
@@ -0,0 +1,275 @@
...
+        }
+      }
+      tb = child.nextReady();
+    }
+
+    /*
+     * We know that child.nextReady() has returned <code>null</code>, so we have processed all tuple we can. Child is
+     * either EOS or we have to wait for more data.
+     */
+    if (child.eos()) {
```
senderista: Why can't you return whatever aggregation results you currently have available? Since you don't persist aggregation results to your `TupleBuffer` until you know the current group key will never be seen again, why can't you finalize the current `TupleBatch` and return that? I think that's what you have to do for this to be called "streaming"...

src/edu/washington/escience/myria/operator/agg/StreamingAggregate.java
```diff
@@ -0,0 +1,275 @@
...
+      fromIndex += aggregators[agg].getResultSchema().numColumns();
+    }
+  }
+
+  /**
+   * @return A batch's worth of result tuples from this aggregate.
+   * @throws DbException if there is an error.
+   */
+  private TupleBatch getResultBatch() throws DbException {
+    Preconditions.checkState(getChild().eos(), "cannot extract results from an aggregate until child has reached EOS");
```
senderista: I don't understand this at all. I thought the whole point of a "streaming aggregator" was that you *could* return aggregation results before the child reached EOS. Am I missing something? (I assume that you're familiar with the difference in semantics between `TupleBuffer` and `TupleBatchBuffer`. I wonder if you're having trouble with the fact that you can't extract a `TupleBatch` from a `TupleBuffer` until it's finalized. If so, the answer is to use `TupleBatchBuffer` unless you need random access to the rows in the buffer.)

src/edu/washington/escience/myria/operator/agg/StreamingAggregate.java
```diff
@@ -0,0 +1,275 @@
...
+      finalBuffer = resultBuffer.finalResult();
+      if (resultBuffer.numTuples() == 0) {
+        throw new DbException("0 tuples in result buffer");
+      }
+      resultBuffer = null;
+    }
+    if (finalBuffer.isEmpty()) {
+      return null;
+    } else {
+      return finalBuffer.get(0);
```
senderista: So this means that you can only return a single `TupleBatch` from your `TupleBuffer` even if you have accumulated multiple `TupleBatch`es?

senderista: If the above is true, then that suggests you need a unit test where you return multiple `TupleBatch`es of aggregation results (e.g. by reading more than one `TupleBatch` of singleton groups).

test/edu/washington/escience/myria/operator/StreamingAggTest.java
```diff
@@ -0,0 +1,1093 @@
...
+      source.putFloat(2, value);
+      source.putLong(3, value);
+      source.putDateTime(4, new DateTime(2010 + value, 1, 1, 0, 0));
+      source.putString(5, "" + value);
+      source.putBoolean(6, (i / (numTuples / 2) == 0));
+      source.putLong(7, 2L);
+    }
+    return source;
+  }
+
```
senderista: As I hinted at earlier, I think you need at least one new unit test to verify the streaming semantics: one which generates more than one `TupleBatch` worth of aggregation results, and one (possibly the same test) which verifies that when multiple `TupleBatch`es of results are generated, the first `TupleBatch` of results is returned before the child is exhausted (i.e., the operator is actually "streaming").

src/edu/washington/escience/myria/operator/agg/StreamingAggregate.java
```diff
@@ -0,0 +1,275 @@
...
+    TupleBatch tb = child.nextReady();
+    while (tb != null) {
+      for (int row = 0; row < tb.numTuples(); ++row) {
+        if (curGroupKey == null) {
+          /* First time accessing this tb, no aggregation performed previously. */
+          // store current group key as a tuple
+          curGroupKey = new Tuple(groupSchema);
+          for (int gKey = 0; gKey < gFields.length; ++gKey) {
+            gTypes[gKey] = tb.getSchema().getColumnType(gFields[gKey]);
+            switch (gTypes[gKey]) {
```
senderista: Can't you skip all this boilerplate by using `TupleUtils.copyValue()`?

src/edu/washington/escience/myria/operator/agg/StreamingAggregate.java
```diff
@@ -0,0 +1,275 @@
...
+                curGroupKey.set(gKey, tb.getDouble(gFields[gKey], row));
+                break;
+            }
+          }
+        } else if (!TupleUtils.tupleEquals(tb, gFields, row, curGroupKey, gRange, 0)) {
+          /* Different grouping key than current one, flush current agg result to result buffer. */
+          addToResult();
+          // store current group key as a tuple
+          for (int gKey = 0; gKey < gFields.length; ++gKey) {
+            switch (gTypes[gKey]) {
```
senderista: Can't you skip all this boilerplate by using `TupleUtils.copyValue()`?

test/edu/washington/escience/myria/operator/StreamingAggTest.java
yqguo: It is added as testMultiBatchResult().

src/edu/washington/escience/myria/operator/agg/StreamingAggregate.java
```diff
@@ -0,0 +1,199 @@
...
+          /* Different grouping key than current one, flush current agg result to result buffer. */
+          addToResult();
+          // store current group key as a tuple
+          for (int gKey = 0; gKey < gFields.length; ++gKey) {
+            TupleUtils.copyValue(tb, gFields[gKey], row, curGroupKey, gKey);
+          }
+          aggregatorStates = AggUtils.allocateAggStates(aggregators);
+          if (resultBuffer.hasFilledTB()) {
+            return resultBuffer.popFilled();
+          }
```
yqguo: `fetchNextReady()` is changed such that as soon as the aggregator has filled a `TupleBatch`, it will return that `TupleBatch` and pause aggregation at that point. The next call to `fetchNextReady()` resumes aggregation until it fills another `TupleBatch`. Unfilled `TupleBatch` is expected only when `child.eos()` is reached.

src/edu/washington/escience/myria/operator/agg/StreamingAggregate.java
```diff
@@ -0,0 +1,199 @@
...
+   * field(s).
+   *
+   * @throws DbException if any error occurs.
+   * @return result tuple batch
+   */
+  @Override
+  protected TupleBatch fetchNextReady() throws DbException {
+    if (child.eos()) {
+      return resultBuffer.popAny();
+    }
```
yqguo: I was thinking to simply `return null` here, because the last result `TupleBatch`, filled or not, is expected to be returned after processing the last input tuple. So here would be the case when `resultBuffer` is empty. But I wasn't sure...

senderista: Yes, you should return `null`. It would be misleading to do otherwise.

src/edu/washington/escience/myria/operator/agg/StreamingAggregate.java
```diff
@@ -0,0 +1,199 @@
...
+    if (child == null) {
+      return null;
+    }
+    Schema inputSchema = child.getSchema();
+    if (inputSchema == null) {
+      return null;
+    }
+
+    groupSchema = inputSchema.getSubSchema(gFields);
+    resultSchema = groupSchema;
```
senderista: This won't work because `groupSchema` is now aliased to `resultSchema`, so you'll be mutating both in the following code. There really should be a `Schema` copy constructor, but instead you could just do
`resultSchema = Schema.of(groupSchema.getColumnTypes(), groupSchema.getColumnNames())`.

src/edu/washington/escience/myria/operator/agg/StreamingAggregate.java
senderista: Looks good!

PR #749, opened 2015-05-22
Sampling Operators

radion:
@senderista Here is my cleaned up version of the sampling operators. The tests pass and I am able to run queries with them. Let me know of any feedback.


radion:
Transient error on the Travis build unrelated to my changes here:

`edu.washington.escience.myria.systemtest.JsonQuerySubmitTest > ingestTest FAILED
      org.junit.runners.model.TestTimedOutException: test timed out after 120000 milliseconds`

radion:
Update: currently investigating an issue that occasionally causes a samplescan query to never terminate.

src/edu/washington/escience/myria/operator/SamplingDistribution.java
```diff
@@ -0,0 +1,300 @@
...
+  private int sampleSize;
+
+  /** True if using a percentage instead of a specific tuple count. */
+  private boolean isPercentageSample = false;
+
+  /** Percentage of total tuples to sample. */
+  private float samplePercentage;
+
+  /** The type of sampling to perform. Currently supports 'WR' and 'WoR'. */
+  private final String sampleType;
```
senderista: I think this should be an enum (looks like Jackson supports serializing enums as strings). There can be compatibility issues with binary serialization of enums as integer values, but I think serializing them as strings should be fine.

src/edu/washington/escience/myria/api/encoding/SamplingDistributionEncoding.java
```diff
@@ -0,0 +1,35 @@
...
+public class SamplingDistributionEncoding extends UnaryOperatorEncoding<SamplingDistribution> {
+
+  /** A specific number of tuples to sample. */
+  public Integer sampleSize;
+
+  /** Percentage of total tuples to sample. */
+  public Float samplePercentage;
+
+  @Required
+  public String sampleType;
```
senderista: See comment in SamplingDistribution.java, I think if you make this an enum, it should "just work" due to default Jackson magic.

src/edu/washington/escience/myria/operator/Sample.java
```diff
@@ -0,0 +1,241 @@
...
+  private int streamSize;
+
+  /** Number of tuples to sample from the right operator. */
+  private int sampleSize;
+
+  /** Random generator used for index selection. */
+  private Random rand;
+
+  /** The type of sampling to perform. Currently supports 'WR' and 'WoR'. */
+  private String sampleType;
```
senderista: See comment in SamplingDistribution.java about this being an enum.

src/edu/washington/escience/myria/operator/Sample.java
```diff
@@ -0,0 +1,241 @@
...
+  private boolean computedSamplingInfo = false;
+
+  /** Buffer for tuples that will be returned. */
+  private TupleBatchBuffer ans;
+
+  /** Global count of the tuples seen so far. */
+  private int tupleNum = 0;
+
+  /** Sorted array of tuple indices that will be taken as samples. */
+  private int[] samples;
```
senderista: I'd prefer a more descriptive name like `sampleIndices`.

src/edu/washington/escience/myria/operator/Sample.java
```diff
@@ -0,0 +1,241 @@
...
+  private String sampleType;
+
+  /** True if operator has extracted sampling info. */
+  private boolean computedSamplingInfo = false;
+
+  /** Buffer for tuples that will be returned. */
+  private TupleBatchBuffer ans;
+
+  /** Global count of the tuples seen so far. */
+  private int tupleNum = 0;
```
senderista: I'd prefer a more descriptive name like `numTuplesSeen` or just `tuplesSeen`.

src/edu/washington/escience/myria/operator/Sample.java
```diff
@@ -0,0 +1,241 @@
...
+    } else if (col2Type == Type.LONG_TYPE) {
+      sampleSize = (int) tb.getLong(2, 0);
+    } else {
+      throw new DbException("SampleSize column must be of type INT or LONG");
+    }
+    Preconditions.checkState(sampleSize >= 0, "sampleSize cannot be negative");
+
+    Type col3Type = tb.getSchema().getColumnType(3);
+    if (col3Type == Type.STRING_TYPE) {
+      sampleType = tb.getString(3, 0);
```
senderista: If we change `SampleType` to an enum then we should catch an exception from parsing an unknown string value and throw the appropriate `DbException`.

src/edu/washington/escience/myria/operator/Sample.java
```diff
@@ -0,0 +1,241 @@
...
+
+    // Return a ready tuple batch if possible.
+    TupleBatch nexttb = ans.popAny();
+    if (nexttb != null) {
+      return nexttb;
+    }
+    // Check if there's nothing left to sample.
+    if (curSampIdx >= samples.length) {
+      getRight().close();
+      setEOS();
```
senderista: This should be unnecessary if both children have been closed.

src/edu/washington/escience/myria/operator/Sample.java
```diff
@@ -0,0 +1,241 @@
...
+    Set<Integer> indices = new HashSet<Integer>(sampleSize);
+    for (int i = populationSize - sampleSize; i < populationSize; i++) {
+      int idx = rand.nextInt(i + 1);
+      if (indices.contains(idx)) {
+        indices.add(i);
+      } else {
+        indices.add(idx);
+      }
+    }
+    int[] indicesArr = new int[indices.size()];
```
senderista: Since we already use Guava, try `Ints.toArray()`.

src/edu/washington/escience/myria/operator/Sample.java
```diff
@@ -0,0 +1,241 @@
...
+   *
+   * @param populationSize
+   *          size of the population that will be sampled from.
+   * @param sampleSize
+   *          number of samples to draw from the population.
+   * @return a sorted array of indices.
+   */
+  private int[] generateIndicesWoR(int populationSize, int sampleSize) {
+    Set<Integer> indices = new HashSet<Integer>(sampleSize);
+    for (int i = populationSize - sampleSize; i < populationSize; i++) {
```
senderista: This looks like Floyd's algorithm, in which case you should give a reference, since it's sufficiently non-obvious to likely cause puzzlement.

src/edu/washington/escience/myria/operator/Sample.java
```diff
@@ -0,0 +1,241 @@
...
+  public Sample(final Operator left, final Operator right, Long randomSeed) {
+    super(left, right);
+    this.rand = new Random();
+    if (randomSeed != null) {
+      this.rand.setSeed(randomSeed);
+    }
+  }
+
+  @Override
+  protected TupleBatch fetchNextReady() throws Exception {
```
senderista: I'm not sure we have to be *this* lazy; this sort of expensive initialization is normally done in `init()`, so it's performed on `open()`. If you don't have a compelling reason for deferring it until `nextReady()` is called, I think it's clearer to put the sample computation in `init()`.

src/edu/washington/escience/myria/operator/Sample.java
```diff
@@ -0,0 +1,241 @@
...
+  private int[] samples;
+
+  /** Current index of the samples array. */
+  private int curSampIdx = 0;
+
+  /**
+   * Instantiate a Sample operator using sampling info from the left operator
+   * and the stream from the right operator.
+   *
+   * @param left
```
senderista: I would be interested in some background on the design decision to represent the sampling parameters as a single tuple from an arbitrary child operator.

src/edu/washington/escience/myria/operator/Sample.java
```diff
@@ -0,0 +1,241 @@
...
+      getRight().close();
+      setEOS();
+      return null;
+    }
+    Operator right = getRight();
+    for (TupleBatch tb = right.nextReady(); tb != null; tb = right.nextReady()) {
+      if (curSampIdx >= samples.length) { // done sampling
+        break;
+      }
+      if (samples[curSampIdx] > tupleNum + tb.numTuples()) {
```
senderista: Shouldn't this be `>=`?

src/edu/washington/escience/myria/operator/Sample.java
```diff
@@ -0,0 +1,241 @@
...
+        // nextIndex is not in this batch. Continue with next batch.
```
senderista: Isn't this short-circuit unnecessary since the while-loop condition will be initially unsatisfied and we increment `tupleNum` and hit the bottom of the loop? The only difference is we don't check for a filled `TupleBatch` in this codepath, which shouldn't matter since we didn't add any results in this iteration.

src/edu/washington/escience/myria/operator/Sample.java
radion: Yes, I do believe you are right...

src/edu/washington/escience/myria/operator/Sample.java
radion: Agreed.

src/edu/washington/escience/myria/operator/SampledDbInsertTemp.java
```diff
@@ -0,0 +1,221 @@
...
+  private MutableTupleBuffer reservoir = null;
+  /** Sampled tuples ready to be returned. */
+  private List<TupleBatch> batches;
+  /** Next element of batches List that will be returned. */
+  private int batchNum = 0;
+  /** True if all samples have been gathered from the child. */
+  private boolean doneSamplingFromChild;
+
+  /** The output schema. */
+  private static final Schema COUNT_SCHEMA = Schema.of(
```
senderista: `Schema.ofFields()` would be cleaner.

src/edu/washington/escience/myria/operator/SampledDbInsertTemp.java
```diff
@@ -0,0 +1,221 @@
...
+   *          the key of the table that the tuple counts will be inserted into.
+   * @param connectionInfo
+   *          the parameters of the database connection.
+   */
+  public SampledDbInsertTemp(final Operator child, final int streamSampleSize,
+      final RelationKey sampleRelationKey, final RelationKey countRelationKey,
+      final ConnectionInfo connectionInfo) {
+    super(child);
+    // Sampling setup.
+    Preconditions.checkArgument(streamSampleSize >= 0L,
```
senderista: Nitpick: `L` qualifier is unnecessary for comparing `int` variable to `int` literal.

src/edu/washington/escience/myria/operator/SampledDbInsertTemp.java
```diff
@@ -0,0 +1,221 @@
...
+        "sampleSize must be non-negative");
+    this.streamSampleSize = streamSampleSize;
+    doneSamplingFromChild = false;
+
+    // Relation setup.
+    Objects.requireNonNull(sampleRelationKey, "sampleRelationKey");
```
senderista: `Objects.requireNonNull()` is designed to be inlined into the assignment (ditto Guava `Preconditions.checkNotNull()`.

src/edu/washington/escience/myria/operator/SampledDbInsertTemp.java
```diff
@@ -0,0 +1,221 @@
...
+      for (int i = 0; i < tb.numTuples(); i++) {
+        if (reservoir.numTuples() < streamSampleSize) {
+          // Reservoir size < k. Add this tuple.
+          for (int j = 0; j < tb.numColumns(); j++) {
+            reservoir.put(j, columns.get(j), i);
+          }
+        } else {
+          // Replace probabilistically
+          int replaceIdx = rand.nextInt(tupleCount);
+          if (replaceIdx < reservoir.numTuples()) {
```
senderista: i think it might be clearer to use `streamSampleSize` instead of `reservoir.numTuples()`, otherwise the reader might get the impression that `reservoir.numTuples()` could be something other than `streamSampleSize` at this point in the code.

src/edu/washington/escience/myria/operator/SampledDbInsertTemp.java
```diff
@@ -0,0 +1,221 @@
...
+   * @throws DbException
+   *           if TupleBatch fails to get nextReady
+   */
+  private void fillReservoir() throws DbException {
+    Random rand = new Random();
+    for (TupleBatch tb = getChild().nextReady(); tb != null; tb = getChild()
+        .nextReady()) {
+      final List<? extends Column<?>> columns = tb.getDataColumns();
+      for (int i = 0; i < tb.numTuples(); i++) {
+        if (reservoir.numTuples() < streamSampleSize) {
```
senderista: Reservoir sampling is much better-known than Floyd's sampling algorithm, but it might still be helpful to add a reference (Wikipedia article is fine), so the reader knows exactly which algorithm is being used.

src/edu/washington/escience/myria/operator/SampledDbInsertTemp.java
```diff
@@ -0,0 +1,221 @@
...
+  public void cleanup() {
+    reservoir = null;
+    batches = null;
+
+    try {
+      if (accessMethod != null) {
+        accessMethod.close();
+      }
+    } catch (DbException e) {
+      throw new RuntimeException(e);
```
senderista: I realize that this is pasted from the other DBInsert operators, but I'm not sure why we need to wrap this in a `RuntimeException`, given that `Operator.close()` can deal with `DbException` being thrown from `cleanup()` (and `cleanup()` explicitly throws `Exception`, so can throw any checked exception). @jingjingwang could you comment?

src/edu/washington/escience/myria/operator/SampledDbInsertTemp.java
```diff
@@ -0,0 +1,221 @@
...
+import edu.washington.escience.myria.column.Column;
+import edu.washington.escience.myria.column.builder.IntColumnBuilder;
+import edu.washington.escience.myria.parallel.RelationWriteMetadata;
+import edu.washington.escience.myria.storage.MutableTupleBuffer;
+import edu.washington.escience.myria.storage.TupleBatch;
+
+/**
+ * Samples the stream into a temp relation.
+ */
+public class SampledDbInsertTemp extends UnaryOperator implements DbWriter {
```
senderista: Why couldn't this class extend `DbInsertTemp`? I think you should be able to override `consumeTuples()` to sample from each `TupleBatch` into the reservoir, and override `childEOS()` to insert the reservoir into the temp table.

src/edu/washington/escience/myria/operator/SampledDbInsertTemp.java
```diff
@@ -0,0 +1,221 @@
...
+      fillReservoir();
+      batches = reservoir.getAll();
+      // Insert sampled tuples into sampleRelationKey
+      while (batchNum < batches.size()) {
+        TupleBatch batch = batches.get(batchNum);
+        accessMethod.tupleBatchInsert(sampleRelationKey, batch);
+        batchNum++;
+      }
+
+      // Write (WorkerID, PartitionSize, StreamSize) to countRelationKey
```
senderista: Shouldn't the `StreamSize` attribute be renamed to something like SampleSize or LocalSampleSize? What does "stream" mean in this context (I thought it referred to all the tuples in the input)?

src/edu/washington/escience/myria/operator/SampledDbInsertTemp.java
```diff
@@ -0,0 +1,221 @@
...
+  public final Schema generateSchema() {
+    if (getChild() == null) {
+      return null;
+    }
+    return getChild().getSchema();
+  }
+
+  @Override
+  public Map<RelationKey, RelationWriteMetadata> writeSet() {
+    Map<RelationKey, RelationWriteMetadata> map = new HashMap<RelationKey, RelationWriteMetadata>(2);
```
senderista: Why not just `ImmutableMap.of()`?

domoritz: Or guava?

src/edu/washington/escience/myria/util/SamplingType.java
```diff
@@ -0,0 +1,9 @@
+package edu.washington.escience.myria.util;
+
+/**
+ * Enumeration of supported sampling types.
+ */
+public enum SamplingType {
+  // WithReplacement, WithoutReplacement
+  WR, WoR
```
senderista: Why not just the long forms of the names? It would be more to type but would be self-documenting...

src/edu/washington/escience/myria/operator/SamplingDistribution.java
```diff
@@ -0,0 +1,320 @@
...
+import edu.washington.escience.myria.column.builder.StringColumnBuilder;
+import edu.washington.escience.myria.storage.TupleBatch;
+import edu.washington.escience.myria.util.SamplingType;
+
+public class SamplingDistribution extends UnaryOperator {
+  /** Required for Java serialization. */
+  private static final long serialVersionUID = 1L;
+
+  /** The output schema. */
+  private static final Schema SCHEMA = Schema.of(ImmutableList.of(
```
senderista: `Schema.ofFields()`?

src/edu/washington/escience/myria/operator/SamplingDistribution.java
```diff
@@ -0,0 +1,320 @@
...
+  }
+
+  @Override
+  public Schema generateSchema() {
+    return SCHEMA;
+  }
+
+  @Override
+  protected void init(final ImmutableMap<String, Object> execEnvVars) {
+    rand = new Random();
```
senderista: As I said above, I think all of this can just be moved to the constructor, since nothing here is expensive and nothing can fail. `tupleCounts` and `streamCounts` could be initialized in their declarations.

src/edu/washington/escience/myria/operator/SamplingDistribution.java
```diff
@@ -0,0 +1,320 @@
...
+  ArrayList<Integer> streamCounts;
+
+  /** Total number of tuples across all workers. */
+  int totalTupleCount = 0;
+
+  private SamplingDistribution(Operator child, SamplingType sampleType,
+      Long randomSeed) {
+    super(child);
+    this.sampleType = sampleType;
+    this.randomSeed = randomSeed;
```
senderista: I think it would be reasonable to put any operations which are not expensive and can't fail into the constructor, which would allow you to eliminate the `randomSeed` instance variable and avoid overriding` init()`. It's a subjective call, though.

src/edu/washington/escience/myria/operator/SamplingDistribution.java
```diff
@@ -0,0 +1,320 @@
...
+  private float samplePercentage;
+
+  /** The type of sampling to perform. */
+  private final SamplingType sampleType;
+
+  /** Random generator used for creating the distribution. */
+  private Random rand;
+
+  /** Seed for the random generator. */
+  private Long randomSeed;
```
senderista: Can be eliminated if you initialize `rand` in the constructor.

src/edu/washington/escience/myria/operator/SamplingDistribution.java
```diff
@@ -0,0 +1,320 @@
...
+  /** Required for Java serialization. */
+  private static final long serialVersionUID = 1L;
+
+  /** The output schema. */
+  private static final Schema SCHEMA = Schema.of(ImmutableList.of(
+      Type.INT_TYPE, Type.INT_TYPE, Type.INT_TYPE, Type.STRING_TYPE),
+      ImmutableList.of("WorkerID", "StreamSize", "SampleSize", "SampleType"));
+
+  /** Total number of tuples to sample. */
+  private int sampleSize;
```
senderista: This will be default-initialized to 0, of course, but given that one of the constructors doesn't explicitly initialize it, and it's only set much later, it might be clearer to initialize it to 0 explicitly.

src/edu/washington/escience/myria/operator/SamplingDistribution.java
```diff
@@ -0,0 +1,320 @@
...
+   *          extracts (WorkerID, PartitionSize, StreamSize) information from
+   *          this child.
+   * @param randomSeed
+   *          value to seed the random generator with. null if no specified seed
+   */
+  public SamplingDistribution(Operator child, int sampleSize,
+      SamplingType sampleType, Long randomSeed) {
+    this(child, sampleType, randomSeed);
+    this.sampleSize = sampleSize;
+    Preconditions.checkState(this.sampleSize >= 0,
```
senderista: `Preconditions.checkArgument()` would be better form.

src/edu/washington/escience/myria/operator/SamplingDistribution.java
```diff
@@ -0,0 +1,320 @@
...
+   *          percentage of total samples to create a distribution for.
+   * @param sampleType
+   *          the type of sampling distribution to create
+   * @param child
+   *          extracts (WorkerID, PartitionSize, StreamSize) information from
+   *          this child.
+   * @param randomSeed
+   *          value to seed the random generator with. null if no specified seed
+   */
+  public SamplingDistribution(Operator child, float samplePercentage,
```
senderista: Is "percentage" necessarily the right interface? It's more natural from the implementation perspective to take a float between 0 and 1, and if the target audience is scientists, it may be more natural for them as well (just as, say, p-values are expressed as real number fractions, not percentages). Additionally, a "percentage" parameter leaves it ambiguous whether only integer values are acceptable (though the float type makes that obvious, but the type isn't visible in JSON query plans). Anyway, just being devil's advocate here, not saying you made the wrong choice.

src/edu/washington/escience/myria/operator/SamplingDistribution.java
```diff
@@ -0,0 +1,320 @@
...
+      SamplingType sampleType, Long randomSeed) {
+    this(child, sampleType, randomSeed);
+    this.isPercentageSample = true;
+    this.samplePercentage = samplePercentage;
+    Preconditions.checkState(samplePercentage >= 0 && samplePercentage <= 100,
```
senderista: `Preconditions.checkArgument()` would be better form.

src/edu/washington/escience/myria/operator/SamplingDistribution.java
```diff
@@ -0,0 +1,320 @@
...
+    // Drain out all the workerID and partitionSize info.
+    while (!getChild().eos()) {
+      TupleBatch tb = getChild().nextReady();
+      if (tb == null) {
+        if (getChild().eos()) {
+          break;
+        }
+        return null;
+      }
+      Type col0Type = tb.getSchema().getColumnType(0);
```
senderista: I think you could factor out the sampling info extraction pretty cleanly into a separate helper (since I think it just mutates instance variables) and make this method more readable.

src/edu/washington/escience/myria/operator/SamplingDistribution.java
```diff
@@ -0,0 +1,320 @@
...
+    if (sampleType == SamplingType.WR) {
+      sampleCounts = withReplacementDistribution(tupleCounts, sampleSize);
+    } else if (sampleType == SamplingType.WoR) {
+      sampleCounts = withoutReplacementDistribution(tupleCounts, sampleSize);
+    } else {
+      throw new DbException("Invalid sampleType: " + sampleType);
+    }
+
+    // Build and return a TupleBatch with the distribution.
+    IntColumnBuilder wIdCol = new IntColumnBuilder();
```
senderista: I think you could save a bit of code by using `TupleBatchBuffer` and `popAny()`.

src/edu/washington/escience/myria/operator/SamplingDistribution.java
```diff
@@ -0,0 +1,320 @@
...
+        if (col0Type == Type.INT_TYPE) {
+          workerID = tb.getInt(0, i);
+        } else if (col0Type == Type.LONG_TYPE) {
+          workerID = (int) tb.getLong(0, i);
+        } else {
+          throw new DbException("WorkerID must be of type INT or LONG");
+        }
+        Preconditions.checkState(workerID > 0, "WorkerID must be > 0");
+        // Ensure the future .set(workerID, -) calls will work.
+        for (int j = tupleCounts.size(); j < workerID; j++) {
```
senderista: Are `workerID`s guaranteed to be consecutive and always begin at 1? If not (and I don't think they will be, since relations are usually not partitioned on all workers), we should use a `Map` instead for `tupleCounts`/`streamCounts`, otherwise you'll have "dummy" entries which could cause bugs.

src/edu/washington/escience/myria/operator/SamplingDistribution.java
```diff
@@ -0,0 +1,320 @@
...
+    for (int val : tupleCounts) {
+      totalTupleCount += val;
+    }
+
+    for (int i = 0; i < sampleSize; i++) {
+      int sampleTupleIdx = rand.nextInt(totalTupleCount);
+      // Assign this tuple to the workerID that holds this sampleTupleIdx.
+      int tupleOffset = 0;
+      for (int j = 0; j < tupleCounts.size(); j++) {
+        if (sampleTupleIdx < tupleCounts.get(j) + tupleOffset) {
```
senderista: I think this code still works if there are "gaps" in the `workerID` sequence (so some `tupleCounts` entries are 0), but that's pretty fragile code. It would be better to iterate through all the existing `workerID` entries in order (say with a `SortedMap`), and avoid these dummy elements altogether.

src/edu/washington/escience/myria/operator/SamplingDistribution.java
```diff
@@ -0,0 +1,320 @@
...
+
+    for (int i = 0; i < sampleSize; i++) {
+      int sampleTupleIdx = rand.nextInt(totalTupleCount - i);
+      // Assign this tuple to the workerID that holds this sampleTupleIdx.
+      int tupleOffset = 0;
+      for (int j = 0; j < logicalTupleCounts.size(); j++) {
+        if (sampleTupleIdx < logicalTupleCounts.get(j) + tupleOffset) {
+          distribution[j] += 1;
+          // Cannot sample the same tuple, so pretend it doesn't exist anymore.
+          logicalTupleCounts.set(j, logicalTupleCounts.get(j) - 1);
```
senderista: Are you sure this logic is correct? It looks suspicious that you're effectively decreasing the population size without decreasing `totalTupleCount`. I'm not even sure it's possible to do this correctly without recording the samples already chosen. In that case you could just use Floyd's algorithm to compute the sample explicitly and calculate the number of samples for each worker at the end. It could be this is obviously correct though, since I'm stupid from painkillers. Do you have a simple proof of correctness?

src/edu/washington/escience/myria/operator/network/partition/IdentityHashPartitionFunction.java
```diff
@@ -0,0 +1,57 @@
...
+  private final int index;
+
+  /**
+   * @param index the index of the partition field.
+   */
+  @JsonCreator
+  public IdentityHashPartitionFunction(
+      @JsonProperty(value = "index", required = true) final Integer index) {
+    super(null);
+    this.index = java.util.Objects.requireNonNull(index, "missing property index");
```
senderista: Why not import `Objects`?

src/edu/washington/escience/myria/operator/network/partition/IdentityHashPartitionFunction.java
```diff
@@ -0,0 +1,57 @@
...
+
+import com.fasterxml.jackson.annotation.JsonCreator;
+import com.fasterxml.jackson.annotation.JsonProperty;
+import com.google.common.base.Preconditions;
+
+import edu.washington.escience.myria.Type;
+import edu.washington.escience.myria.storage.TupleBatch;
+
+/**
+ * Implementation of a PartitionFunction that use the trivial identity hash.
```
senderista: Could you make it explicit that the relation must have an `INT` column that stores `workerID`s?

src/edu/washington/escience/myria/operator/SampledDbInsertTemp.java
radion: Agreed. Will refactor.

src/edu/washington/escience/myria/operator/SampledDbInsertTemp.java
radion: The naming convention comes from the need for SamplingDistribution to know how many tuples were 'pre-sampled' from this stream. I will change the name to PartitionSampleSize to hopefully remove the confusion here.

src/edu/washington/escience/myria/operator/SamplingDistribution.java
radion: I do intentionally create the dummy entries, which means that those workers will be told to sample 0 tuples.

src/edu/washington/escience/myria/operator/SamplingDistribution.java
radion: I can describe the detailed decision here in person sometime, but the main idea is that it's simulating the removal of the tuple from the population. I don't have a formal proof that this is legitimate to do, but it does pass all the tests I've thrown at it :)

src/edu/washington/escience/myria/operator/SamplingDistribution.java
radion: I see your point here. Part of the reasoning here is to make it easier for RACO to know the difference between specifying tupleSize vs percentSize. I think I'll leave it as-is for the time-being but a good discussion to have at some point. (it'll be straightforward to modify the code)

src/edu/washington/escience/myria/operator/network/partition/IdentityHashPartitionFunction.java
radion: Done.

src/edu/washington/escience/myria/operator/Sample.java
radion: Either way, it shouldn't really change the correctness or runtime of sampling. I'll think about this for a future commit.

src/edu/washington/escience/myria/operator/SamplingDistribution.java
```diff
@@ -0,0 +1,328 @@
...
+  private float samplePercentage;
+
+  /** The type of sampling to perform. */
+  private final SamplingType sampleType;
+
+  /** Random generator used for creating the distribution. */
+  private Random rand;
+
+  /** Maps (worker_i) --> (sampling info for worker_i) */
+  TreeMap<Integer, WorkerInfo> workerInfo = new TreeMap<>();
```
senderista: Total nitpick, but this should be declared as `SortedMap` (the interface).

src/edu/washington/escience/myria/operator/SamplingDistribution.java
```diff
@@ -0,0 +1,328 @@
...
+   *
+   * @param workerInfo
+   *          reference to the workerInfo to modify.
+   * @param totalTupleCount
+   *          total # of tuples across all workers.
+   * @param sampleSize
+   *          total # of samples to distribute across the workers.
+   */
+  private void withReplacementDistribution(
+      TreeMap<Integer, WorkerInfo> workerInfo, int totalTupleCount,
```
senderista: `SortedMap`

src/edu/washington/escience/myria/operator/SamplingDistribution.java
```diff
@@ -0,0 +1,328 @@
...
+   *
+   * @param workerInfo
+   *          reference to the workerInfo to modify.
+   * @param totalTupleCount
+   *          total # of tuples across all workers.
+   * @param sampleSize
+   *          total # of samples to distribute across the workers.
+   */
+  private void withoutReplacementDistribution(
+      TreeMap<Integer, WorkerInfo> workerInfo, int totalTupleCount,
```
senderista: `SortedMap`

src/edu/washington/escience/myria/operator/SamplingDistribution.java
```diff
@@ -0,0 +1,328 @@
...
+      int sampleSize) {
+    Map<Integer, Integer> logicalTupleCounts = new TreeMap<>();
```
senderista: If you're depending on ordered iteration of keys, I think this should be declared as `SortedMap`, even if you only use the `Map` interface.

src/edu/washington/escience/myria/operator/SamplingDistribution.java
```diff
@@ -0,0 +1,328 @@
...
+  }
+
+  @Override
+  protected TupleBatch fetchNextReady() throws DbException {
+    // Drain out all the worker info.
+    while (!getChild().eos()) {
+      TupleBatch tb = getChild().nextReady();
+      if (tb == null) {
+        if (getChild().eos()) {
+          break;
```
senderista: Should we eagerly close the child here?

src/edu/washington/escience/myria/operator/SamplingDistribution.java
```diff
@@ -0,0 +1,328 @@
...
+      withReplacementDistribution(workerInfo, totalTupleCount, sampleSize);
+    } else if (sampleType == SamplingType.WithoutReplacement) {
+      withoutReplacementDistribution(workerInfo, totalTupleCount, sampleSize);
+    } else {
+      throw new DbException("Invalid sampleType: " + sampleType);
+    }
+
+    // Build and return a TupleBatch with the distribution.
+    // Assumes that the sampling information can fit into one tuple batch.
+    IntColumnBuilder wIDs = new IntColumnBuilder();
```
senderista: You could also use `ColumnFactory.allocateColumns(SCHEMA)` to create the list of `ColumnBuilder`s in one line, although you might sacrifice some explicitness unless you named the column indices.

src/edu/washington/escience/myria/operator/SamplingDistribution.java
```diff
@@ -0,0 +1,328 @@
...
+  private void withoutReplacementDistribution(
+      TreeMap<Integer, WorkerInfo> workerInfo, int totalTupleCount,
+      int sampleSize) {
+    Map<Integer, Integer> logicalTupleCounts = new TreeMap<>();
+    for (Map.Entry<Integer, WorkerInfo> wInfo : workerInfo.entrySet()) {
+      logicalTupleCounts.put(wInfo.getKey(), wInfo.getValue().tupleCount);
+    }
+
+    for (int i = 0; i < sampleSize; i++) {
+      int sampleTupleIdx = rand.nextInt(totalTupleCount - i);
```
senderista: So I guess this is kind of a modified truncated Fisher-Yates shuffle, right? I worry that even if your trick of decrementing  a worker partition each time it gets assigned a sampled index is correct, you are still invalidating the assumptions of the Fisher-Yates algorithm. The standard behavior is to swap the value at the sampled index with the value at `length - i - 1`, so the value just sampled can never be sampled again, and the value at `length - i - 1` *can* be sampled again. But you aren't doing any swapping, logical or otherwise (because you don't materialize the array and you aren't retaining any sampled values), so how do you expect to preserve the correctness of the algorithm?

senderista: Never mind, I think I've convinced myself your algorithm is correct :)

src/edu/washington/escience/myria/operator/SamplingDistribution.java
```diff
@@ -0,0 +1,328 @@
...
+      }
+
+      WorkerInfo wInfo = new WorkerInfo(tupleCount, actualTupleCount);
+      workerInfo.put(workerID, wInfo);
+      totalTupleCount += tupleCount;
+    }
+  }
+
+  /**
+   * Creates a WithoutReplacement distribution across the workers.
```
senderista: typo

src/edu/washington/escience/myria/operator/SampledDbInsertTemp.java
```diff
@@ -0,0 +1,156 @@
...
+import edu.washington.escience.myria.column.Column;
+import edu.washington.escience.myria.column.builder.IntColumnBuilder;
+import edu.washington.escience.myria.parallel.RelationWriteMetadata;
+import edu.washington.escience.myria.storage.MutableTupleBuffer;
+import edu.washington.escience.myria.storage.TupleBatch;
+
+/**
+ * Samples the stream into a temp relation.
+ */
+public class SampledDbInsertTemp extends DbInsertTemp implements DbWriter {
```
senderista: Nitpick: the interface `DbWriter` is already implemented by the grandparent (`AbstractDBInsert`), so is redundant here.

PR #751, opened 2015-06-03
Implement catalog scan

domoritz:
Fixes https://github.com/uwescience/myria/issues/747


domoritz:
@mbalazin @jingjingwang @senderista Who want's to review this PR?

senderista:
I could probably take a look tomorrow.

src/edu/washington/escience/myria/accessmethod/SQLiteTupleBatchIterator.java
```diff
@@ -0,0 +1,104 @@
...
+    final boolean hasRow = statement.hasRow();
+    if (!hasRow) {
+      statement.dispose();
+      connection.dispose();
+    }
+    return hasRow;
+  }
+
+  @Override
+  public TupleBatch next() {
```
senderista: I realize that this you're just moving this class into a new file, but this implementation incorrectly assumes the client always calls `hasNext()` before `next()`. It appears that we will just return an empty `TupleBatch` indefinitely if the client keeps calling `next()` after results have been exhausted (and the statement/connection won't be cleaned up). Conceptually, if `next()` is called and `hasNext()` would return `false`, we should throw `NoSuchElementException`, per the documented semantics for `Iterator` (http://docs.oracle.com/javase/7/docs/api/java/util/Iterator.html#next()). This should be as simple as calling `hasNext()` at the beginning of `next()` and throwing `NoSuchElementException` if it returns `false`.

src/edu/washington/escience/myria/coordinator/catalog/MasterCatalog.java
```diff
@@ -1983,4 +1984,33 @@ protected String job(final SQLiteConnection sqliteConnection) throws CatalogExce
...
+  public Iterator<TupleBatch> tupleBatchIteratorFromQuery(final String queryString, final Schema outputSchema)
+      throws CatalogException {
+    try {
+      return queue.execute(new SQLiteJob<Iterator<TupleBatch>>() {
+        @Override
+        protected Iterator<TupleBatch> job(final SQLiteConnection sqliteConnection) throws CatalogException,
+            SQLiteException {
+          SQLiteStatement statement = sqliteConnection.prepare(queryString);
+          List<TupleBatch> tuples = Lists.newLinkedList();
+          Iterator<TupleBatch> iter = new SQLiteTupleBatchIterator(statement, sqliteConnection, outputSchema);
```
senderista: Why not just `Iterators.addAll(tuples, new SQLiteTupleBatchIterator(statement, sqliteConnection, outputSchema))`?

src/edu/washington/escience/myria/operator/CatalogQueryScan.java
```diff
@@ -0,0 +1,94 @@
...
+import com.google.common.collect.ImmutableMap;
+
+import edu.washington.escience.myria.DbException;
+import edu.washington.escience.myria.Schema;
+import edu.washington.escience.myria.coordinator.catalog.CatalogException;
+import edu.washington.escience.myria.coordinator.catalog.MasterCatalog;
+import edu.washington.escience.myria.storage.TupleBatch;
+
+/**
+ * Push a select query down into a JDBC based database and scan over the query result.
```
senderista: This description should be made specific to the catalog.

src/edu/washington/escience/myria/operator/CatalogQueryScan.java
```diff
@@ -0,0 +1,94 @@
...
+
+  /**
+   * Constructor.
+   * 
+   * @param sql see the corresponding field.
+   * @param outputSchema see the corresponding field.
+   * @param catalog see the corresponding field.
+   * */
+  public CatalogQueryScan(final String sql, final Schema outputSchema, final MasterCatalog catalog) {
+    Objects.requireNonNull(sql);
```
senderista: You can inline all these calls to `requireNonNull()` into the assignments.

src/edu/washington/escience/myria/operator/CatalogQueryScan.java
```diff
@@ -0,0 +1,94 @@
...
+    Objects.requireNonNull(outputSchema);
+    Objects.requireNonNull(catalog);
+
+    this.sql = sql;
+    this.outputSchema = outputSchema;
+    this.catalog = catalog;
+    tuples = null;
```
senderista: If you think you need this superfluous initialization for explicitness (of course it will be default-initialized to `null`), I think it should be in the declaration.

src/edu/washington/escience/myria/operator/CatalogQueryScan.java
```diff
@@ -0,0 +1,94 @@
...
+    }
+  }
+
+  @Override
+  public final Schema generateSchema() {
+    return outputSchema;
+  }
+
+  @Override
+  protected final void init(final ImmutableMap<String, Object> execEnvVars) throws DbException {
```
senderista: Why implement an empty override that's not abstract?

src/edu/washington/escience/myria/operator/CatalogQueryScan.java
```diff
@@ -0,0 +1,94 @@
...
+      } catch (CatalogException e) {
+        throw new DbException(e);
+      }
+    }
+    if (tuples.hasNext()) {
+      final TupleBatch tb = tuples.next();
+      LOGGER.trace("Got {} tuples", tb.numTuples());
+      return tb;
+    } else {
+      return null;
```
senderista: Shouldn't you `setEOS()` here (since you don't have a child it won't propagate from the child)?

test/edu/washington/escience/myria/operator/CatalogScanTest.java
```diff
@@ -0,0 +1,75 @@
...
+   * Destroy catalog.
+   */
+  @After
+  public void Cleanup() {
+    catalog.close();
+  }
+
+  @Test
+  public final void testQueryQueries() throws DbException, CatalogException {
+    Schema schema = new Schema(ImmutableList.of(Type.LONG_TYPE, Type.STRING_TYPE), ImmutableList.of("id", "raw"));
```
senderista: `Schema.ofFields()`, or at least `Schema.of()` (we need to do something about public exposure of the `Schema` constructor).

src/edu/washington/escience/myria/accessmethod/SQLiteTupleBatchIterator.java
domoritz: Good catch. I don't see why somebody would call next without hasNext but with this fix we definitely avoid infinite loops.

src/edu/washington/escience/myria/coordinator/catalog/MasterCatalog.java
domoritz: That was what I was looking for. I tried `tuples.addAll` but t didn't work because iter is not a collection. Thanks!

src/edu/washington/escience/myria/operator/CatalogQueryScan.java
domoritz: I don't think so. At least my test doesn't fail and it tests for EOS.

domoritz:
@senderista All done.

PR #752, opened 2015-06-13
Refactor deployment

jingjingwang:
This PR does several things:

1. Remove worker catalogs. Worker configuration values are added to the "runtime" section of their deployment config files. If not specified, we used the default values in Java files.
2. Change the directory hierarchy of deployments. The old deployment path of a worker is: path/name-files/name/worker, with "path" and "name" specified in a deployment.cfg. The new path is: path/name/worker_x for worker x and path/name/master for master. Worker/master stdout/err, worker.cfg, master.catalog are put inside of their directories. 
3. With these changes, a lot of deployment scripts can be removed/merged. Now we have only one script setup_cluster.py, with an option "-fresh" if we want to start with a new master catalog. Otherwise the master catalog at the deployment path is not changed. Worker config files are always recreated and synced based on the deployment.cfg so that everything is up-to-date after running setup_cluster.py.

Sorry for putting almost everything inside of one commit, since I found it to be hard to separate them into commits that can compile and pass tests.


jingjingwang:
@senderista There are changes affecting production so I assigned it to you. 
@BrandonHaynes You may also want to check it out :)

BrandonHaynes:
Is the `[runtime]` section optional such that existing deployment configuration files will be backwards-compatible, or this now a required section?  Also, are there any optional keys inside the `[runtime]` section besides `max_heap_size` that I should know about?  I'll get the deployment creation utility updated.

BrandonHaynes:
Nit: "reinitialize" or "overwrite" might be a bit more intuitive option name than "fresh" in `setup_cluster.py`.

jingjingwang:
Yes it's an optional section. Heap size is probably the only thing, other keys/values can be specified but usually not used. I changed the arg to be "-fresh_catalog" to be specific.

myriadeploy/deployment.cfg.local
```diff
@@ -16,3 +14,8 @@ rest_port = 8753
 [workers]
 1 = localhost:9001
 2 = localhost:9002
+
+[runtime]
+# Uncomment to set the maximum heap size of the worker processes
+#max_heap_size = -Xmx2g
```
senderista: I guess we need to retain backward compatibility, but why do we need to use exactly the format of the JVM command-line option for heap size? It doesn't make any sense to someone unfamiliar with the JVM command line. How about a name like `max_heap_size_gb` so the value could be a simple number?

myriadeploy/get_logs.py
```diff
@@ -15,11 +15,10 @@ def get_std_logs_from_worker(hostname, dirname, username,
                              worker_id, description):
     mkdir_if_not_exists(description)
     if hostname == 'localhost':
-        uri = "%s/worker_%s_stdout" % (dirname, worker_id)
+        uri = "%s/worker_%s/worker_%s_stdout" % (dirname, worker_id, worker_id)
```
senderista: Another convention to consider is `workers/1/` etc. This is less uniform WRT `master/`, but might make scenarios like running `grep -R` over all worker files simpler.

myriadeploy/get_logs.py
```diff
@@ -28,30 +27,29 @@ def get_error_logs_from_worker(hostname, dirname, username,
                                worker_id, description):
     mkdir_if_not_exists(description)
     if hostname == 'localhost':
-        uri = "%s/worker_%s_stderr" % (dirname, worker_id)
+        uri = "%s/worker_%s/worker_%s_stderr" % (dirname, worker_id, worker_id)
```
senderista: Why do we need to qualify the name of files under a worker-specific directory with the worker ID?

myriadeploy/get_logs.py
```diff
@@ -28,30 +27,29 @@ def get_error_logs_from_worker(hostname, dirname, username,
...
+            username, hostname, dirname, worker_id, worker_id)
+    args = ["scp", uri, "%s/worker_%s_stderr" % (description, worker_id,)]
     return subprocess.call(args)
 
 
 def get_logs_from_master(hostname, dirname, username, description):
     mkdir_if_not_exists(description)
     if hostname == 'localhost':
-        uri = "%s/master_stdout" % (dirname)
+        uri = "%s/master/master_stdout" % (dirname)
```
senderista: Same question as above.

myriadeploy/deployment.cfg.local
BrandonHaynes: I almost suggested this too -- but if you do go this route, it might make sense to expose an additional "JVM flags" key since I know that people use this to push other command-line options.

myriadeploy/setup_cluster.py
```diff
@@ -2,55 +2,30 @@
...
-    if subprocess.call(args):
-        logging.error("Error copying worker catalogs")
-
-
-def copy_distribution(config_file):
+def make_deployment(config_file, fresh):
     "Copy the distribution (jar and libs and conf) to compute nodes."
-    args = ["./using_deployment_utils.sh", config_file, "-copy_distribution"]
+    args = ["./using_deployment_utils.sh", config_file, "-deploy"]
+    if fresh:
```
senderista: I'm not sure that "fresh" has a lot of usage in CS outside macro bindings. Maybe "clean" (as in `make clean`, "clean install", etc.) would be more familiar/suggestive?

myriadeploy/setup_cluster.py
```diff
@@ -2,55 +2,30 @@
...
     if subprocess.call(args):
         logging.error("Error copying distribution")
 
 
 def main(argv):
     # Usage
-    if len(argv) != 2:
-        print >> sys.stderr, "Usage: %s <deployment.cfg>" % (argv[0])
+    if len(argv) < 2 or len(argv) > 3 or (len(argv) == 3 and argv[2] != "-fresh_catalog"):
+        print >> sys.stderr, "Usage: %s <deployment.cfg> <optional: -fresh_catalog>" % (argv[0])
```
senderista: In most Unix programs, options precede ordinary arguments and (outside Java-land) are in short form (one letter, with a single-hyphen prefix) or long form (one or more hyphen-separated complete words, with a two-hyphen prefix).

myriadeploy/setup_cluster.py
```diff
@@ -2,55 +2,30 @@
...
         print >> sys.stderr, \
             "       deployment.cfg: \
             a configuration file modeled after deployment.cfg.sample"
+        print >> sys.stderr, \
+            "       -fresh_catalog: \
+            if deploy with an empty master catalog"
```
senderista: "deployed" or "deploying"

myriadeploy/setup_cluster.py
BrandonHaynes: Ideally would switch to `argparse` here, which does all of this for you.  (Q.v. `create_deployment.py`)

speedtest/edu/washington/escience/myria/mrbenchmarks/Main.java
```diff
@@ -103,10 +94,14 @@ public static void main(String[] args) throws Exception {
...
-    for (Entry<Integer, SocketInfo> e : allWorkerInfos.entrySet()) {
-      workerName2ID.put(e.getValue().getHost(), e.getKey());
+    MyriaConfiguration config =
+        MyriaConfiguration.loadWithDefaultValues(FilenameUtils.concat(masterHome, MyriaConstants.DEPLOYMENT_CONF_FILE));
+    for (String id : config.getWorkerIds()) {
+      workerName2ID.put(config.getHostname(id), Integer.parseInt(id));
     }
 
+    BufferedReader br =
+        new BufferedReader(new InputStreamReader(new FileInputStream(new File(queryWorkerNameListFile))));
```
senderista: Why not just `BufferedReader(FileReader(String))`?

speedtest/edu/washington/escience/myria/sp2bench/Main.java
```diff
@@ -7,24 +7,24 @@
...
 import edu.washington.escience.myria.parallel.QueryFuture;
 import edu.washington.escience.myria.parallel.Server;
 import edu.washington.escience.myria.storage.TupleBatch;
+import edu.washington.escience.myria.tool.MyriaConfiguration;
 import edu.washington.escience.myria.util.DateTimeUtils;
 
 public class Main {
 
-  final static String catalogFileName = FilenameUtils.concat("/tmp/slxu_experiment", "master.catalog");
+  final static String masterHome = "/tmp/slxu_experiment";
```
senderista: Could we have a more descriptive path name?

src/edu/washington/escience/myria/tool/MyriaConfiguration.java
```diff
@@ -0,0 +1,261 @@
+package edu.washington.escience.myria.tool;
```
senderista: "tools" would be a more natural and accurate name than "tool".

test/edu/washington/escience/myria/operator/CatalogScanTest.java
```diff
@@ -28,7 +28,7 @@
   /**
    * The catalog
    */
-  private MasterCatalog catalog;
+  private edu.washington.escience.myria.coordinator.MasterCatalog catalog;
```
senderista: Why not use the import?

speedtest/edu/washington/escience/myria/sqlite/TwitterJoinSpeedTest.java
```diff
@@ -58,7 +59,8 @@ public void before() {
       final Path src = FileSystems.getDefault().getPath(srcPath[i]);
       final Path dst =
           FileSystems.getDefault().getPath(
-              FilenameUtils.concat(getWorkerFolder(workerIDs[i]), "worker_" + workerIDs[i] + "_data.db"));
+              FilenameUtils.concat(DeploymentUtils.getPathToWorkerDir(workingDir, workerIDs[i]), "worker_"
```
senderista: Another example of how duplicating the worker ID in file names requires unnecessary code.

speedtest/edu/washington/escience/myria/sp2bench/Main.java
```diff
@@ -85,16 +83,15 @@ public static void main(String[] args) throws Exception {
...
-    catalog.close();
-
-    int[] allWorkers = new int[workerSet.size()];
+    MyriaConfiguration config =
+        MyriaConfiguration.loadWithDefaultValues(FilenameUtils.concat(masterHome, MyriaConstants.DEPLOYMENT_CONF_FILE));
+    int[] allWorkers = new int[config.getWorkerIds().size()];
     int idx = 0;
-    for (int wID : workerSet) {
-      allWorkers[idx++] = wID;
+    for (String id : config.getWorkerIds()) {
```
senderista: Why doesn't `getWorkerIds()` return a list of `int`s?

src/edu/washington/escience/myria/MyriaSystemConfigKeys.java
```diff
@@ -57,98 +56,76 @@ private MyriaSystemConfigKeys() {
    * */
   public static final String TCP_CONNECTION_TIMEOUT_MILLIS = "tcp.connection.timeout.milliseconds";
```
senderista: I guess if you're gonna replace the Java-style dotted names with underscores (unless there's implicit hierarchy), you should do it everywhere.

src/edu/washington/escience/myria/MyriaSystemConfigKeys.java
```diff
@@ -57,98 +56,76 @@ private MyriaSystemConfigKeys() {
...
-   * Add default configurations into a configuraion.
+   * Add default configurations into a configuration.
    * 
    * @param config the configuration.
    * */
-  public static void addDefaultConfigKeys(final Map<String, String> config) {
-    if (!config.containsKey(FLOW_CONTROL_WRITE_BUFFER_HIGH_MARK_BYTES)
-        || config.get(FLOW_CONTROL_WRITE_BUFFER_HIGH_MARK_BYTES) == null) {
-      config.put(FLOW_CONTROL_WRITE_BUFFER_HIGH_MARK_BYTES,
+  public static void addDefaultConfigValues(final MyriaConfiguration config) {
```
senderista: Could we factor out all this default-setting into a helper? E.g., `setDefaultValueIfMissing(String section, String key, String defaultValue)`

src/edu/washington/escience/myria/coordinator/ConfigFileException.java
```diff
@@ -0,0 +1,21 @@
+/**
```
senderista: Delete this empty comment block?

src/edu/washington/escience/myria/coordinator/ConfigFileGenerator.java
```diff
@@ -0,0 +1,90 @@
...
+ * 
+ */
+public final class ConfigFileGenerator {
+
+  /**
+   * @param args args[0]: config file. args[1]: optional, location to store config files.
+   * @throws ConfigFileException if there is an error creating the config files.
+   */
+  public static void main(final String[] args) throws ConfigFileException {
+    Preconditions.checkArgument(args.length == 2, "Usage: ConfigFileGenerator config_file location");
```
senderista: Make it clearer that the first argument is a file name?

src/edu/washington/escience/myria/coordinator/ConfigFileGenerator.java
```diff
@@ -0,0 +1,90 @@
...
+
+  /**
+   * Creates a worker config file.
+   * 
+   * @param config the parsed configuration.
+   * @param workerId the worker whose catalog is being creating.
+   * @param path the place to save the worker config file.
+   * @throws ConfigFileException error when creating the config file.
+   */
+  public static File makeOneWorkerConfigFile(final MyriaConfiguration config, final String workerId, final String path)
```
senderista: Why can't we strongly type `workerId`?

src/edu/washington/escience/myria/tool/MyriaConfiguration.java
```diff
@@ -0,0 +1,261 @@
...
+    return getHostname(nodeId) + ":" + getPort(nodeId);
+  }
+
+  /**
+   * 
+   * @param nodeId the worker/master ID
+   * @return the port number
+   * @throws ConfigFileException if error occurred when getting the value
+   */
+  public String getPort(final String nodeId) throws ConfigFileException {
```
senderista: Why can't this be numerically typed?

src/edu/washington/escience/myria/tool/MyriaConfiguration.java
```diff
@@ -0,0 +1,261 @@
...
+      return getRequired("workers", nodeId).split(":")[1];
+    }
+  }
+
+  /**
+   * 
+   * @return a list of worker IDs
+   * @throws ConfigFileException if error occurred when getting the value
+   */
+  public List<String> getWorkerIds() throws ConfigFileException {
```
senderista: Why can't this be numerically typed?

src/edu/washington/escience/myria/tool/MyriaConfiguration.java
```diff
@@ -0,0 +1,261 @@
...
+    return FilenameUtils.concat(getPath(nodeId), getRequired("deployment", MyriaSystemConfigKeys.DESCRIPTION));
+  }
+
+  /**
+   * 
+   * @param workerId the worker ID
+   * @return the database name
+   * @throws ConfigFileException if error occurred when getting the value
+   */
+  public String getWorkerDatabaseName(final String workerId) throws ConfigFileException {
```
senderista: Why can't `workerId` be numerically typed?

src/edu/washington/escience/myria/parallel/Server.java
```diff
@@ -412,36 +412,26 @@ QueryExecutionMode getExecutionMode() {
   /**
    * Construct a server object, with configuration stored in the specified catalog file.
    * 
-   * @param catalogFileName the name of the file containing the catalog.
+   * @param configFile the name of the config file.
    * @throws FileNotFoundException the specified file not found.
    * @throws CatalogException if there is an error reading from the Catalog.
```
senderista: Delete this @throws comment?

src/edu/washington/escience/myria/parallel/Server.java
```diff
@@ -412,36 +412,26 @@ QueryExecutionMode getExecutionMode() {
...
-    /* Get the master socket info */
-    List<SocketInfo> masters = catalog.getMasters();
-    if (masters.size() != 1) {
-      throw new RuntimeException("Unexpected number of masters: expected 1, got " + masters.size());
-    }
-    masterSocketInfo = masters.get(MyriaConstants.MASTER_ID);
-
-    workers = new ConcurrentHashMap<>(catalog.getWorkers());
-    final ImmutableMap<String, String> allConfigurations = catalog.getAllConfigurations();
+  public Server(final String configFile) throws FileNotFoundException, CatalogException, ConfigFileException {
```
senderista: Remove `CatalogException` from `throws` clause?

src/edu/washington/escience/myria/storage/Tuple.java
```diff
@@ -157,7 +159,7 @@ public void putBoolean(final int column, final boolean value) {
   }
 
   @Override
-  public void putDateTime(final int column, final DateTime value) {
+  public void putDateTime(final int column, @Nonnull final DateTime value) {
```
senderista: Is there any reason to not add the same annotation to `putString` and `putObject`?

src/edu/washington/escience/myria/tool/MyriaConfiguration.java
```diff
@@ -0,0 +1,261 @@
...
+  /**
+   * load the config file.
+   * 
+   * @param filename filename.
+   * @return parsed mapping from sections to keys to values.
+   * @throws ConfigFileException if error occurred when parsing the config file
+   * */
+  public static MyriaConfiguration loadWithDefaultValues(final String filename) throws ConfigFileException {
+    File f = new File(filename);
+    if (!f.exists()) {
```
senderista: Why not just let a `FileNotFoundException` be thrown later when you try to read the file?

src/edu/washington/escience/myria/tool/MyriaConfiguration.java
```diff
@@ -0,0 +1,261 @@
...
+      throw new RuntimeException("config file " + filename + " doesn't exist!");
+    }
+    MyriaConfiguration config = new MyriaConfiguration();
+    try {
+      config.read(f);
+      MyriaSystemConfigKeys.addDefaultConfigValues(config);
```
senderista: I don't think this line can fail, so it shouldn't be in the try block.

src/edu/washington/escience/myria/tool/MyriaConfiguration.java
```diff
@@ -0,0 +1,261 @@
...
+    } catch (IOException e) {
```
senderista: `addDefaultConfigValues` can't throw `IOException`, right?

systemtest/edu/washington/escience/myria/daemon/MasterDaemonTest.java
```diff
@@ -146,29 +139,50 @@ public void testStartAndRestShutdown() throws Exception {
...
     Response response = invocation.get();
     assertEquals(Status.OK.getStatusCode(), response.getStatus());
-    String deploymentFile = md.getClusterMaster().getConfiguration(MyriaSystemConfigKeys.DEPLOYMENT_FILE);
     String respond = response.readEntity(String.class);
-    /*
-     * Tests and cluster deployments have different working directory hierarchies (no description-files). So here we
-     * only compare the suffix. TODO: unify these two.
-     */
-    assertTrue(respond.endsWith(deploymentFile));
+    assertTrue(respond.equals(Paths.get(workingDir, MyriaConstants.DEPLOYMENT_CONF_FILE).toString()));
```
senderista: `assertEquals`

test/edu/washington/escience/myria/operator/CatalogScanTest.java
senderista: Why not use the import?

src/edu/washington/escience/myria/util/DeploymentUtils.java
```diff
@@ -158,30 +81,118 @@ public static void main(final String[] args) throws IOException {
...
+    startMaster(hostname, workingDir, maxHeapSize, restPort, ssl);
+  }
+
+  /**
+   * 
+   * @param config a Myria configuration.
+   * @param workerId worker ID.
+   * @throws ConfigFileException if error occurred parsing config file.
+   */
+  public static void startAWorker(final MyriaConfiguration config, final String workerId) throws ConfigFileException {
```
senderista: Why can't this just be called `startWorker` since you've already overloaded `startMaster`?

src/edu/washington/escience/myria/util/DeploymentUtils.java
```diff
@@ -158,30 +81,118 @@ public static void main(final String[] args) throws IOException {
...
+   * 
+   * @param configFile the path to a Myria config file.
+   * @return the temp deployment directory.
+   * @throws IOException I/O exception.
+   */
+  public static File createTempDeployment(final String configFile) throws IOException {
+    final String current = System.getProperty("user.dir");
+    File tempDeploy = Files.createTempDirectory("tempdeploy").toFile();
+    String tempDeployPath = tempDeploy.getAbsolutePath();
+    System.out.println("generating temp deployment at " + tempDeployPath);
```
senderista: Delete this line?

src/edu/washington/escience/myria/util/DeploymentUtils.java
```diff
@@ -158,30 +81,118 @@ public static void main(final String[] args) throws IOException {
...
+    List<String> includes = Arrays.asList("worker_" + workerId);
+    List<String> excludes = Arrays.asList("worker_*", "master");
+    rsyncFileToRemote(localDeployPath + "/", hostname, workingDir, includes, excludes);
+  }
+
+  /**
+   * 
+   * @param localDeployPath source local deployment.
+   * @param config a Myria configuration.
+   * @param freshCatalog if deploy with an empty master catalog.
```
senderista: "deployed"

src/edu/washington/escience/myria/util/DeploymentUtils.java
```diff
@@ -158,30 +81,118 @@ public static void main(final String[] args) throws IOException {
...
+    System.out.println("Start syncing distribution files to worker#" + workerId + " @ " + hostname);
+    mkdir(hostname, workingDir);
+    List<String> includes = Arrays.asList("worker_" + workerId);
+    List<String> excludes = Arrays.asList("worker_*", "master");
+    rsyncFileToRemote(localDeployPath + "/", hostname, workingDir, includes, excludes);
+  }
+
+  /**
+   * 
+   * @param localDeployPath source local deployment.
```
senderista: "source of"

src/edu/washington/escience/myria/util/DeploymentUtils.java
```diff
@@ -158,30 +81,118 @@ public static void main(final String[] args) throws IOException {
...
-      final String maxHeapSize, final String workerId, final int port, final boolean debug) {
+  private static void startWorker(final String address, final String workingDir, final String maxHeapSize,
+      final String workerId, final int port, final boolean debug) {
     StringBuilder builder = new StringBuilder();
-    String path = workingDir + "/" + description + "-files";
-    String workerDir = path + "/" + description + "/" + "worker_" + workerId;
-    String classpath = "'" + path + "/conf:" + path + "/libs/*'";
-    String librarypath = path + "/" + "sqlite4java-392";
+    String workerDir = workingDir + "/" + "worker_" + workerId;
+    String classpath = "'" + workingDir + "/conf:" + workingDir + "/libs/*'";
```
senderista: This is kind of a mess and could be cleaned up a bit with `String.format`.

src/edu/washington/escience/myria/util/DeploymentUtils.java
```diff
@@ -205,8 +216,8 @@ public static void startWorker(final String address, final String workingDir, fi
     builder.append(" edu.washington.escience.myria.parallel.Worker");
     builder.append(" --workingDir " + workerDir);
     builder.append(" 0</dev/null");
-    builder.append(" 1>" + path + "/worker_" + workerId + "_stdout");
-    builder.append(" 2>" + path + "/worker_" + workerId + "_stderr");
+    builder.append(" 1>" + workerDir + "/worker_" + workerId + "_stdout");
```
senderista: Again I think this could be simplified by not duplicating `workerID` in filenames.

src/edu/washington/escience/myria/util/DeploymentUtils.java
```diff
@@ -335,24 +349,30 @@ public static void mkdir(final String address, final String remotePath) {
    * @param remotePath e.g. /tmp/test
    */
   public static void rsyncFileToRemote(final String localPath, final String address, final String remotePath) {
-    rsyncFileToRemote(localPath, address, remotePath, false);
+    rsyncFileToRemote(localPath, address, remotePath, null, null);
```
senderista: This can't be correct given the code below (which doesn't tolerate `null` for `includes` or `excludes`).

src/edu/washington/escience/myria/util/DeploymentUtils.java
```diff
@@ -335,24 +349,30 @@ public static void mkdir(final String address, final String remotePath) {
...
-      final boolean useDel) {
+      final List<String> includes, final List<String> excludes) {
     ArrayList<String> command = new ArrayList<String>();
     command.add("rsync");
     command.add("-rtLDvz");
-    if (useDel) {
-      command.add("--del");
+    /* Remove old files at dest. */
+    command.add("--del");
+    for (String inc : includes) {
```
senderista: This will certainly cause an NPE if `includes` or `excludes` is `null`. Was this codepath tested? Personally I favor empty collections over `null` (simpler code), so I think you should leave the code as-is and change the callers to supply empty collections in place of `null`.

myriadeploy/deployment.cfg.local
jingjingwang: Thanks for both suggestions, since we don't have other JVM flags to apply right now, I changed it to be "max_heap_size_gb" which takes a number, and also add "min_heap_size_gb". If we have more in the future, adding a "JVM flags" key can be a solution.

myriadeploy/setup_cluster.py
jingjingwang: I adopted `argparse`, the code looks much cleaner now. Thanks for the suggestion!

speedtest/edu/washington/escience/myria/mrbenchmarks/Main.java
jingjingwang: I simply moved this line from above to this place with no change. Yes it should be `BufferedReader(FileReader(String))`, fixed.

src/edu/washington/escience/myria/coordinator/ConfigFileGenerator.java
jingjingwang: I realized that this main method is never used, so I removed it.

src/edu/washington/escience/myria/storage/Tuple.java
jingjingwang: Simply because I didn't see them in the list of Java warnings when fixing `putDateTime` :) Fixed.

src/edu/washington/escience/myria/util/DeploymentUtils.java
jingjingwang: I realized that this method is never used, so removed. I also added `@Nonnull` to `includes` and `excludes`.

jingjingwang:
All comments above have been addressed, most in 0aaf5de.

BrandonHaynes:
One potential issue here -- for deployment on the MIT HPC, I need to set some environment flags (e.g., `-Djava.net.preferIPv4Stack=true`).  My current approach was to hide these in the heap line of the deployment file.  I haven't looked, but I think I can hack around this with something like `max_heap_size_gb=2gb -Djava.net.preferIPv4Stack=true`.  So, while these workarounds are all hacks, but there exists at least one use case for "extra" flags beyond heap size.  

No need to deal with that for this PR (so long as the workaround continues to workaround) but we may want to extend this to support additional flags.  I'm happy to look into doing so, since I have the need.

jingjingwang:
I see, it's indeed good to have a key for Java flags then. We already have a list of `-Dxxx` in `DeploymentUtils.java` when launching master/workers, maybe we can make them use the same interface. I'll leave it to another PR though.
BTW we changed `max_heap_size_gb` to take a number so it's gonna be `max_heap_size_gb=2`.

senderista:
Per discussion with @BrandonHaynes, we will support arbitrary JVM flags in this PR. Additionally, the `max_heap_size_gb` key will be changed to `heap.size.max.gb`.

BrandonHaynes:
`jvm.heap.size.max.gb`?  I'll probably want a `jvm` (`java`?) prefix for the arbitrary flags.

jingjingwang:
`jvm` sounds better.

jingjingwang:
OK, I think all the comments have been addressed for this PR.

myriadeploy/create_deployment.py
```diff
@@ -129,8 +138,7 @@ def main(argv):
                            '(default is [--name])')
 
     parser.add_argument(
-        '--heap', type=str, help='Java VM heap size (e.g., "-Xmx2g") '
-                                 'and/or other parameters')
+        '--heap', type=float, help='Java VM heap size in GB (e.g., "-Xmx2g")')
```
senderista: This parameter should be changed to something more descriptive, e.g., `--max-heap-size-gb`. Also the sample usage is now wrong (the parenthesized part should just be deleted).

myriadeploy/create_deployment.py
```diff
@@ -53,6 +52,16 @@ def _get_coordinator(hostname, port=9001):
            '0 = {}:{}\n\n'.format(hostname, port)
 
 
+def _get_runtime(heap=None):
+    """ Generates the runtime section of a Myria deployment file """
+    runtime = '[runtime]\n'
+    if heap:
+        runtime += 'max_heap_size.gb = %s\n' % heap
```
senderista: This should now be `jvm.heap.size.max.gb`, right?

myriadeploy/myriadeploy.py
```diff
@@ -68,9 +68,14 @@ def split_hostportpathdbname_key_append_id(hostport):
     ret['nodes'] = [ret['master']] + ret['workers']
     # .. max_heap_size is the Java maximum heap size
     try:
-        ret['max_heap_size'] = config.get('deployment', 'max_heap_size')
+        ret['max_heap_size.gb'] = config.get('runtime', 'max_heap_size.gb')
```
senderista: `jvm.heap.size.max.gb`

myriadeploy/myriadeploy.py
```diff
@@ -68,9 +68,14 @@ def split_hostportpathdbname_key_append_id(hostport):
...
     except ConfigParser.NoOptionError:
-        ret['max_heap_size'] = ''
+        ret['max_heap_size.gb'] = ''
```
senderista: `jvm.heap.size.max.gb`

myriadeploy/myriadeploy.py
```diff
@@ -68,9 +68,14 @@ def split_hostportpathdbname_key_append_id(hostport):
...
+    # .. min_heap_size is the Java minimum heap size
+    try:
+        ret['min_heap_size.gb'] = config.get('runtime', 'min_heap_size.gb')
```
senderista: `jvm.heap.size.min.gb`

myriadeploy/myriadeploy.py
```diff
@@ -68,9 +68,14 @@ def split_hostportpathdbname_key_append_id(hostport):
...
+    except ConfigParser.NoOptionError:
+        ret['min_heap_size.gb'] = ''
```
senderista: `jvm.heap.size.min.gb`

jingjingwang:
No I didn't ... Thanks! That being said, It would be nice to have some tests for this script. @BrandonHaynes could you check if your stuffs still work with this PR?

BrandonHaynes:
Changes look fine to me.  

I'm happy to set up tests for this (it's probably also worth eventually using `create_deployment.py` to create the default `deploy.*` scripts that come with Myria) -- however, I'm not sure how/if Travis supports multiple languages.

senderista:
How is the test coverage on this (ad hoc is fine for now, I just want to be sure everything works before I merge)?

jingjingwang:
All tests have passed, are you looking for more tests..?

senderista:
I was wondering if the modified deployment scripts (or scripts which exercise the modified deployment code) have been tested.

senderista:
Since the ad hoc testing has been successful, I'll merge this PR and create a separate issue for deployment tests.

BrandonHaynes:
Myria-EC2 updated in https://github.com/uwescience/myria-ec2/pull/17

PR #755, opened 2015-06-30
Fixing deletion from catalog during overwrite.

senderista:
I had to make this fix on the fly during deployment because all sample queries from myria-web were succeeding but would not show results. The issue was a design oversight in the "delete relation" code: it always checked for the `is_deleted` flag to be set before deleting a relation from the catalog, but that only makes sense if we're going through the deletion API. In the overwrite scenario we shouldn't check the flag at all before deleting the relation.


senderista:
I like that idea!

senderista:
Created #757 to address schema issues.

PR #760, opened 2015-08-10
insert to db, add delimiter, fix opId

jingjingwang:
Some json queries are not up-to-date.


senderista:
Is it possible to express the connected components/transitive closure queries as Datalog if not MyriaL? Or is the syntax still up in the air?

jingjingwang:
My changes of expressing them in MyriaL, not Datalog, have not been submitted yet. The syntax may need some modifications, but they work at least on my branch as I demoed a few weeks ago. It's definitely possible to express them in Datalog but I guess the consensus is to first implement MyriaL support for them since MyriaL has more exposure.

senderista:
So could you add a Datalog query to `rawQuery`? It would be helpful for me at least to be able to see at a glance what's going on.

senderista:
LGTM (I don't understand the CC example well enough to comment on the Datalog query, although I am familiar with the label propagation algorithm that it seems to be using).

jingjingwang:
Yes that's exactly how the CC query works.

PR #768, opened 2015-09-11
JVM metrics for Ganglia

senderista:
Currently in production, so you can go to http://dbserver20.cs.washington.edu/ganglia/ and take a look :)

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -148,7 +148,7 @@ dependencies {
   // Expression compiler
   compile "org.codehaus.janino:janino:2.7.7"
 
-  compile "org.postgresql:postgresql:9.3-1102-jdbc41"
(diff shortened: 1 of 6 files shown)


jingjingwang:
This is really cool! Just one question: I tried ganglia web a little bit but couldn't see JVM metrics being displayed. For example, I got "MyriaWorker_Memory_Heap_used: down". Does it mean that once this feature is pushed into production we will be able to see them?

senderista:
It's in production but still in development :) I'm not sure what you're referring to; even when workers are down I can still see their data. Could you send me a URL?

senderista:
LGTM (note that with my current code, the default port 8649 will be used if just a host is specified but not a port).

senderista:
@jingjingwang are you OK to merge?

jingjingwang:
Yep!

PR #772, opened 2015-09-15
fix jxm symlink overwrite.

jingjingwang:
if we ran two `./gradlew jar` without cleaning, it would fail to create the symlink because the link already exists.


senderista:
LGTM

jingjingwang:
Then I'm gonna click the button :)

PR #774, opened 2015-09-18
clean assigned workers before instantiating.

jingjingwang:
The fragment may have been instantiated in a previous iteration with relations having different partitioning schemata that the current iteration caused by overwritting. Not cleaning up the assigned worker set gives us false conflicts.


jingjingwang:
Fixes https://rest.myria.cs.washington.edu:1776/query/query-78340. @shbae FYI :)

senderista:
LGTM

PR #775, opened 2015-09-19
Idb store and schema inference

jingjingwang:


src/edu/washington/escience/myria/api/encoding/IDBControllerEncoding.java
```diff
@@ -34,8 +36,13 @@
 
   @Override
   public IDBController construct(final ConstructArgs args) {
-    return new IDBController(argSelfIdbId, realEosControllerOperatorId, realEosControllerWorkerId, null, null, null,
-        argState.construct(), MoreObjects.firstNonNull(sync, Boolean.FALSE));
+    IDBController controller =
+        new IDBController(argSelfIdbId, realEosControllerOperatorId, realEosControllerWorkerId, null, null, null,
+            argState.construct(), MoreObjects.firstNonNull(sync, Boolean.FALSE));
+    if (relationKey != null) {
+      controller.setStoreRelationKey(relationKey);
```
senderista: Why can't `relationKey` be supplied in the constructor if it isn't set afterward? Having setters that can be invoked after the constructor returns is less safe. If you feel the constructor args are getting out of control, consider using a builder.

src/edu/washington/escience/myria/api/encoding/QueryConstruct.java
```diff
@@ -452,41 +455,47 @@ private static void assignWorkersToFragments(final List<PlanFragmentEncoding> fr
   }
 
   /**
+   * Instantiate operators in the given fragment.
+   * 
+   * @param planFragment the encoded plan fragment.
+   * @param args args
+   * @param allOperators a map to keep instantiated operators.
+   */
+  private static void instantiateOperators(final PlanFragmentEncoding planFragment, final ConstructArgs args,
```
senderista: Might be clearer to call this `instantiateOperatorsFromFragment` or `instantiateOperatorsForFragment` or `instantiateFragmentOperators`?

src/edu/washington/escience/myria/api/encoding/QueryConstruct.java
```diff
@@ -85,18 +85,21 @@
...
 
-    Map<Integer, SubQueryPlan> plan = Maps.newHashMap();
-    Map<PlanFragmentEncoding, RootOperator> instantiatedFragments = Maps.newHashMap();
     Map<Integer, Operator> allOperators = Maps.newHashMap();
     for (PlanFragmentEncoding fragment : fragments) {
-      RootOperator op =
-          instantiateFragment(fragment, args, instantiatedFragments, op2OwnerFragmentMapping, allOperators);
+      instantiateOperators(fragment, args, allOperators);
+    }
+    while (!setConsumerSchema(fragments, allOperators)) {
```
senderista: How hard would it be to trigger an infinite loop here?

src/edu/washington/escience/myria/api/encoding/QueryConstruct.java
```diff
@@ -495,61 +504,47 @@ private static RootOperator instantiateFragment(final PlanFragmentEncoding planF
...
     return fragmentRoot;
   }
 
   /**
+   * 
+   * @param fragments the JSON-encoded query fragments to be executed in parallel
+   * @param allOperators a map to keep instantiated operators.
+   * @return if all the consumers have their schemata set.
+   */
+  private static boolean setConsumerSchema(final List<PlanFragmentEncoding> fragments,
```
senderista: Why does this need to be called in a loop? There must be some kind of recursive propagation that I don't understand.

src/edu/washington/escience/myria/api/encoding/QueryConstruct.java
```diff
@@ -85,18 +85,21 @@
...
+    }
+    Map<Integer, SubQueryPlan> plan = Maps.newHashMap();
+    for (PlanFragmentEncoding fragment : fragments) {
+      RootOperator op = instantiateFragment(fragment, args, allOperators);
       for (Integer worker : fragment.workers) {
-        SubQueryPlan workerPlan = plan.get(worker);
-        if (workerPlan == null) {
-          workerPlan = new SubQueryPlan();
-          plan.put(worker, workerPlan);
+        if (!plan.containsKey(worker)) {
```
senderista: Shame `putIfAbsent()` isn't part of `Map` in Java 7 :(

src/edu/washington/escience/myria/operator/IDBController.java
```diff
@@ -23,7 +32,7 @@
  * Together with the EOSController, the IDBController controls what to serve into an iteration and when to stop an
```
senderista: For code monkeys like me, might be good to clarify what "IDB" means...

senderista: I mean, in the code, not in the PR :)

src/edu/washington/escience/myria/operator/IDBController.java
```diff
@@ -309,6 +330,31 @@ public final void init(final ImmutableMap<String, Object> execEnvVars) throws Db
...
+      if (connectionInfo == null) {
+        throw new DbException("Unable to instantiate DbInsert: connection information unknown");
+      }
+      if (connectionInfo instanceof SQLiteInfo) {
+        /* Set WAL in the beginning. */
+        final File dbFile = new File(((SQLiteInfo) connectionInfo).getDatabaseFilename());
+        SQLiteConnection conn = new SQLiteConnection(dbFile);
+        try {
+          conn.open(true);
+          conn.exec("PRAGMA journal_mode=WAL;");
```
senderista: I don't see the need for this. WAL mode is persistent, so it should only need to be supplied at DB creation.

src/edu/washington/escience/myria/operator/IDBController.java
```diff
@@ -353,4 +405,20 @@ public void setStreamingState(final StreamingState state) {
   public StreamingState getStreamingState() {
     return state;
   }
+
+  /**
+   * 
+   * @param key the relation key to store the final state to
+   */
+  public void setStoreRelationKey(final RelationKey key) {
```
senderista: Per above, can this be moved to the constructor?

src/edu/washington/escience/myria/operator/SymmetricHashJoin.java
```diff
@@ -331,7 +331,13 @@ public SymmetricHashJoin(final List<String> outputColumns, final Operator left,
   @Override
   protected Schema generateSchema() {
     final Schema leftSchema = getLeft().getSchema();
+    if (leftSchema == null) {
```
senderista: What's the new case being handled here? Is this the only operator that needs to handle this case?

src/edu/washington/escience/myria/api/encoding/QueryConstruct.java
senderista: Oh, is the schema propagating from producer to consumer, and then that consumer is a producer for another consumer...?

src/edu/washington/escience/myria/api/encoding/IDBControllerEncoding.java
jingjingwang: Because in some cases we don't need to store the states to disk, i.e. there are multiple IDBs in a query but only one serves as the final query result.

src/edu/washington/escience/myria/api/encoding/QueryConstruct.java
jingjingwang: Yes recursive propagation is possible, e.g.
Consumer_B -> IDBController -> Producer_A
Consumer_A -> IDBController -> Producer_B
If we need to know the schema of Consumer_A, we need the schema of Producer_A then Consumer_B then Producer_B. In this case since an IDBController has two children and one is its initial input which usually has a fixed schema, we won't have infinite loop, but you're right that in general it's possible to construct a plan with infinite loop. I'll address it in the following commits.

src/edu/washington/escience/myria/api/encoding/QueryConstruct.java
jingjingwang: Exactly!

src/edu/washington/escience/myria/operator/IDBController.java
jingjingwang: Discussed above.

src/edu/washington/escience/myria/operator/SymmetricHashJoin.java
jingjingwang: The situation is, it's possible to get `null` here if this join depends on, for example, a consumer that depends on a producer which doesn't have its schema to be set yet. Then we will have `NullPointerException` in this line following:
`Type leftType = leftSchema.getColumnType(leftIndex);`

src/edu/washington/escience/myria/api/encoding/QueryConstruct.java
jingjingwang: Discussed below.

src/edu/washington/escience/myria/operator/IDBController.java
jingjingwang: Great suggestion! I updated this piece of Java doc.

src/edu/washington/escience/myria/operator/IDBController.java
jingjingwang: I copied the same piece of code from `DbInsert` to here: https://github.com/uwescience/myria/blob/master/src/edu/washington/escience/myria/operator/DbInsert.java#L187. Do you think it's not necessary on the `DbInsert

src/edu/washington/escience/myria/api/encoding/QueryConstruct.java
jingjingwang: Yes! I renamed it to be `instantiateFragmentOperators`.

src/edu/washington/escience/myria/operator/IDBController.java
senderista: Correct, it's not necessary there either (it only makes sense in the catalog creation code). (Despite the fact that it's set in these places, the current catalog actually wasn't in WAL mode and I had to set it administratively, which was necessary to enable online backups.)

src/edu/washington/escience/myria/operator/SymmetricHashJoin.java
senderista: OK, did you only modify SymmetricHashJoin because that's the only streaming join operator we have? Could other operators have children with dynamic schema in the future?

src/edu/washington/escience/myria/api/encoding/QueryConstruct.java
```diff
@@ -85,18 +85,21 @@
...
 
-    Map<Integer, SubQueryPlan> plan = Maps.newHashMap();
-    Map<PlanFragmentEncoding, RootOperator> instantiatedFragments = Maps.newHashMap();
     Map<Integer, Operator> allOperators = Maps.newHashMap();
     for (PlanFragmentEncoding fragment : fragments) {
-      RootOperator op =
-          instantiateFragment(fragment, args, instantiatedFragments, op2OwnerFragmentMapping, allOperators);
+      instantiateFragmentOperators(fragment, args, allOperators);
+    }
+    while (setConsumerSchema(fragments, allOperators)) {
```
senderista: I'm still concerned about infinite loop bugs. Could we set an arbitrary "big enough" iteration limit and either assert on it or log an error and throw an exception? It's not elegant but could save a lot of debugging.

src/edu/washington/escience/myria/operator/SymmetricHashJoin.java
jingjingwang: No, it's okay for `generateSchema()` to return `null`, e.g. https://github.com/uwescience/myria/blob/master/src/edu/washington/escience/myria/operator/Apply.java#L124. I modified `SymmetricHashJoin` only because of the potential `NullPointerException`. I haven't noticed any other operator with the same problem.

src/edu/washington/escience/myria/api/encoding/QueryConstruct.java
jingjingwang: Added in 2b76c6b, although it should not happen because once `setConsumerSchema()` returns `false`, we exit.

src/edu/washington/escience/myria/operator/IDBController.java
jingjingwang: Sounds good, I removed both pieces in 2b76c6b.

jingjingwang:
Hmm, turned out we can't remove WAL otherwise tests will fail! E.g. `SQLiteAccessMethod.concurrentlyReadAndWriteTest()`. I'm putting them back.

senderista:
Ah, ok, that's because the test is creating a new SQLite db from scratch, and the first connection to set the journaling mode will determine the persistent mode. Concurrent reads and writes are not possible under the default journaling mode (DELETE), so the test fails when the write can't get an exclusive lock on the db.

senderista:
@jingjingwang: could you rebase the changes for #776 on master and force-push?

senderista:
@jingjingwang: ping

PR #776, opened 2015-09-21
Add count filter

jingjingwang:
A CountFilter StreamingState keeps distinct tuples with their counts and only emits a tuple at the first time when its count hits the threshold. It's designed for serving an IDB with `count` as the aggregate function and a `>= threshold` condition.


jingjingwang:
Merge after https://github.com/uwescience/myria/pull/775.

senderista:
There seems to be a lot of duplication between this PR and #775. Could you rebase and force-push after #775 is merged? I'll review it then.

jingjingwang:
Yes, that's the plan: first get #775 merged.

jingjingwang:
@senderista bump :)

jingjingwang:
@senderista I think this PR is finally ready to be reviewed!

src/edu/washington/escience/myria/api/encoding/QueryConstruct.java
```diff
@@ -97,6 +97,7 @@
        * setting the threshold to be the number of operators is enough.
        */
     }
+    Preconditions.checkArgument(loopCount < allOperators.size());
```
senderista: If the preceding loop terminates when `loopCount == allOperators.size()`, won't the post-increment violate this invariant (actually I think it will make `loopCount == allOperators.size()+1`)? Also note that `setConsumerSchema()` could be called `allOperators.size()+1` times, since the second part of the conjunction is evaluated after the first part. This could be fixed by just swapping the two parts of the conjunction to force short-circuiting.

src/edu/washington/escience/myria/operator/CountFilter.java
```diff
@@ -0,0 +1,214 @@
+package edu.washington.escience.myria.operator;
+
+import java.util.ArrayList;
+import java.util.Arrays;
+import java.util.BitSet;
+import java.util.List;
+
+import jersey.repackaged.com.google.common.base.Preconditions;
```
senderista: wrong import

src/edu/washington/escience/myria/operator/CountFilter.java
```diff
@@ -0,0 +1,214 @@
...
+      }
+      doCount.found = false;
+      doCount.meet = false;
+      doCount.sourceRow = i;
+      tupleIndexList.forEach(doCount);
+      if (!doCount.found) {
+        for (int j : keyColIndices) {
+          uniqueTuples.put(j, columns.get(j), i);
+        }
+        tupleCounts.put(0, new IntArrayColumn(new int[] { 1 }, 1), 0);
```
senderista: Why can't you just use `putInt()` here and not construct this bogus singleton column?

src/edu/washington/escience/myria/operator/CountFilter.java
```diff
@@ -0,0 +1,214 @@
...
+    /** if meet the threshold for the first time. */
+    private boolean meet;
+
+    @Override
+    public void value(final int destRow) {
+      if (TupleUtils.tupleEquals(inputTB, keyColIndices, sourceRow, uniqueTuples, destRow)) {
+        found = true;
+        int oldcount = tupleCounts.getInt(0, destRow);
+        if (oldcount < threshold) {
+          tupleCounts.replace(0, destRow, new IntArrayColumn(new int[] { oldcount + 1 }, 1), 0);
```
senderista: Looks like `MutableTupleBuffer` needs `replaceInt()` etc., analogous to `putInt()` etc.

src/edu/washington/escience/myria/operator/CountFilter.java
```diff
@@ -0,0 +1,214 @@
...
+    Schema schema = getOp().getInputSchema();
+    if (schema == null) {
+      return null;
+    }
+    return schema.getSubSchema(keyColIndices);
+  }
+
+  @Override
+  public void init(final ImmutableMap<String, Object> execEnvVars) {
+    uniqueTupleIndices = new IntObjectHashMap<IntArrayList>();
```
senderista: I know this is how the other hash aggregators do things, but I wonder what the perf impact (cache misses/RAM footprint/GC) of `IntObjectMap` vs. `IntIntMap` is here. Given that the hash function is `MurmurHash3_128`, the collision probability is negligible (you would need ~`2^64` entries before the first expected collision), so I wonder what harm there is in assuming that collisions never happen?

senderista: Eh, I see you're only storing 4 bytes of the 16-byte hash code. This could be upgraded to 8 bytes (`LongIntMap`), which should be enough to prevent collisions before about `2^32` entries. Does that seem safe enough? Even if we're doubling the key size, wouldn't this still be lower-overhead than storing an `IntArrayList` for every hash code? I guess we could use `BigInteger` as a hash key, but that seems like it would be very high-overhead and defeat the whole optimization...

src/edu/washington/escience/myria/operator/CountFilter.java
```diff
@@ -0,0 +1,214 @@
...
+    uniqueTuples = new MutableTupleBuffer(getSchema());
+    tupleCounts =
+        new MutableTupleBuffer(Schema.of(Arrays.asList(new Type[] { Type.INT_TYPE }), Arrays
```
senderista: `Schema.of()` can also take a `varargs` form, with interleaved types and names:
```
Schema.of(Type.INT_TYPE, "count");
```

src/edu/washington/escience/myria/operator/CountFilter.java
```diff
@@ -0,0 +1,214 @@
...
+  @Override
+  public int numTuples() {
+    if (uniqueTuples == null) {
+      return 0;
+    }
+    return uniqueTuples.numTuples();
+  }
+
+  @Override
+  public StreamingState newInstanceFromMyself() {
```
senderista: I know it's part of an existing interface, but this is an awful name! What's wrong with `copy()` or `clone()` (I guess the latter has the baggage of the `Cloneable` interface)?

src/edu/washington/escience/myria/operator/network/Producer.java
```diff
@@ -242,50 +251,57 @@ public void triggered(final IPCEvent event) {
...
-  /** set backup buffers as SimpleAppender. */
-  public void setBackupBufferAsAppender() {
-    triedToSendTuples = new ArrayList<StreamingState>();
+  /**
+   * set backup buffers.
+   * 
+   * @param stateEncoding the streaming state encoding.
+   * 
+   * */
+  public void setBackupBuffer(final StreamingStateEncoding<?> stateEncoding) {
```
senderista: Is it possible to extend `StreamingStateEncoding` with an abstract `getStreamingState()` method that delegates constructing the appropriate `StreamingState` to the corresponding `StreamingStateEncoding` subclass?

systemtest/edu/washington/escience/myria/systemtest/JsonQuerySubmitTest.java
```diff
@@ -205,58 +210,66 @@ public void jsonQuerySubmitTest() throws Exception {
...
-    conn = JsonAPIUtils.ingestData("localhost", masterDaemonPort, ingestA0);
-    if (null != conn.getErrorStream()) {
-      throw new IllegalStateException(getContents(conn));
+  public void jsonQuerySubmitWithResultCheck(final String[] ingests, final String[] queries, final String[] expected,
+      final String[] stored) throws Exception {
+    Preconditions.checkArgument(expected.length == stored.length);
+    for (String ingest : ingests) {
+      File ingestFile = new File(ingest);
+      HttpURLConnection conn = JsonAPIUtils.ingestData("localhost", masterDaemonPort, ingestFile);
+      if (null != conn.getErrorStream()) {
```
senderista: I know you didn't introduce this, but I don't think the constant-first comparison idiom is one we want to encourage.

systemtest/edu/washington/escience/myria/systemtest/JsonQuerySubmitTest.java
```diff
@@ -205,58 +210,66 @@ public void jsonQuerySubmitTest() throws Exception {
...
+    for (String query : queries) {
+      File queryJson = new File(query);
+      HttpURLConnection conn = JsonAPIUtils.submitQuery("localhost", masterDaemonPort, queryJson);
+      if (null != conn.getErrorStream()) {
+        throw new IllegalStateException(getContents(conn));
+      }
+      assertEquals(HttpURLConnection.HTTP_ACCEPTED, conn.getResponseCode());
+      long queryId = getQueryStatus(conn).queryId;
+      conn.disconnect();
+      while (!server.getQueryManager().queryCompleted(queryId)) {
```
senderista: I don't really see the point of using both the JSON API and the in-process `QueryManager` API. If we're going to use the latter, then it would be simpler to wait on a `QueryFuture` rather than polling like this. If we're actually testing the JSON API, then we shouldn't cheat by using the `QueryManager` API.

test/edu/washington/escience/myria/operator/KeepAndSortOnMinTest.java
```diff
@@ -48,7 +48,7 @@ public void testKeepAndSortedOnMinValue() throws DbException {
     while (!keepmin.eos()) {
       keepmin.nextReady();
       long lastValue = -1;
-      List<TupleBatch> result = keepmin.getStreamingState().exportState();
+      List<TupleBatch> result = keepmin.getStreamingStates().get(0).exportState();
```
senderista: Is the change to `getStreamingStates()` solely for the benefit of `Producer`? If so, and if none of the other `StreamingStateful` implementations can logically have more than one `StreamingState`, maybe we should include a convenience `getStreamingState()` method in the interface? Which callers actually expect a list from this method, instead of immediately calling `get(0)` on the result? It seems like every call site I see in this diff is made more awkward by this change.

src/edu/washington/escience/myria/operator/CountFilter.java
senderista: You can ignore these comments for now; I opened https://github.com/uwescience/myria/issues/787 to address this issue.

jingjingwang:
I have address all comments in baa35db, except system tests since they're removed. (Hope we can get them back soon). Sorry for this long delay!

src/edu/washington/escience/myria/operator/CountFilter.java
```diff
@@ -0,0 +1,213 @@
...
+    return schema.getSubSchema(keyColIndices);
+  }
+
+  @Override
+  public void init(final ImmutableMap<String, Object> execEnvVars) {
+    uniqueTupleIndices = new IntObjectHashMap<IntArrayList>();
+    uniqueTuples = new MutableTupleBuffer(getSchema());
+    tupleCounts =
+        new MutableTupleBuffer(
+            Schema.of(ImmutableList.of(Type.INT_TYPE), ImmutableList.of("count")));
```
senderista: `Schema.ofFields(Type.INT_TYPE, "count")`

src/edu/washington/escience/myria/operator/CountFilter.java
```diff
@@ -0,0 +1,213 @@
...
+    /** if meet the threshold for the first time. */
+    private boolean meet;
+
+    @Override
+    public void value(final int destRow) {
+      if (TupleUtils.tupleEquals(inputTB, keyColIndices, sourceRow, uniqueTuples, destRow)) {
+        found = true;
+        int oldcount = tupleCounts.getInt(0, destRow);
+        if (oldcount < threshold) {
+          tupleCounts.replace(0, destRow, new IntArrayColumn(new int[] {oldcount + 1}, 1), 0);
```
senderista: This is a really awkward interface; I wonder how many other examples use a dummy 1-element array like this. Worth adding a new overload of `MutableTupleBuffer.replace()`?

senderista:
Aside from nitpicks above, LGTM

jingjingwang:
Done, I ended up refactoring some methods in `MutableTupleBuffer'.

src/edu/washington/escience/myria/storage/MutableTupleBuffer.java
```diff
@@ -413,6 +322,84 @@ public final void swap(final int column, final int destRow, final int sourceRow)
...
+    }
+  }
+
+  /**
+   * @param destColumn the destination column.
+   * @param destRow the row.
+   * @param value the replacement.
+   */
+  public final void replaceInt(final int destColumn, final int destRow, final int value) {
+    getColumn(destColumn, destRow).replaceInt(value, destRow);
```
senderista: I'm confused why this isn't (here and below):
```
getColumn(destColumn, destRow).replaceInt(value, destRow % TupleBatch.BATCH_SIZE);
```

src/edu/washington/escience/myria/operator/CountFilter.java
```diff
@@ -191,8 +187,9 @@ public void value(final int destRow) {
         found = true;
         int oldcount = tupleCounts.getInt(0, destRow);
         if (oldcount < threshold) {
-          tupleCounts.replace(0, destRow, new IntArrayColumn(new int[] {oldcount + 1}, 1), 0);
-          meet = (oldcount + 1 >= threshold);
+          oldcount += 1;
+          tupleCounts.replaceInt(0, destRow, oldcount);
```
senderista: I believe this won't work unless the new `MutableTupleBuffer.replaceInt()` method is revised to use `destRow % TupleBatch.BATCH_SIZE` as row offset in the column.

src/edu/washington/escience/myria/operator/InMemoryOrderBy.java
```diff
@@ -193,8 +192,7 @@ public void sort() {
       for (int columnIdx = 0; columnIdx < getSchema().numColumns(); columnIdx++) {
         int sourceRow = indexes.get(rowIdx);
         int tupleIdx = table.getTupleIndexInContainingTB(sourceRow);
-        ReadableColumn hashTblColumn = table.getColumns(sourceRow)[columnIdx];
-        ans.put(columnIdx, hashTblColumn, tupleIdx);
+        ans.put(columnIdx, table.getColumn(columnIdx, sourceRow), tupleIdx);
```
senderista: This usage suggests that the `TupleBatchBuffer`/`MutableTupleBuffer` API still isn't high-level enough: an ordinary client shouldn't have to worry about finding the correct `TupleBatch`, extracting the correct column from it, and calculating the correct row offset within that column. They should only have to specify source/destination row/column offsets and values. I think a symptom of this is that `getTupleIndexInContainingTB()` is a public API, which I think shouldn't be necessary (the implementations should be handling that calculation themselves).

senderista:
Refactoring is a big improvement, but as I mentioned in comments, I think `TupleBatchBuffer.put()` still needs some work.

src/edu/washington/escience/myria/operator/CountFilter.java
```diff
@@ -0,0 +1,209 @@
...
+    /** if meet the threshold for the first time. */
+    private boolean meet;
+
+    @Override
+    public void value(final int destRow) {
+      if (TupleUtils.tupleEquals(inputTB, keyColIndices, sourceRow, uniqueTuples, destRow)) {
+        found = true;
+        int newcount = tupleCounts.getInt(0, destRow) + 1;
+        if (newcount <= threshold) {
+          tupleCounts.replaceInt(0, destRow, newcount);
```
senderista: I believe this won't work unless the new `MutableTupleBuffer.replaceInt()` method is revised to use `destRow % TupleBatch.BATCH_SIZE` as row offset in the column.

jingjingwang: You're right, that's a silly mistake.

jingjingwang:
I did more refactoring, now low-level APIs are hided. Callers only need to provide the row/col indices of the source `TupleBatch` or `MutableTupleBuffer` to append a value to a `TupleBatchBuffer`.

senderista:
Love the new `append` methods, they make client code much cleaner! This is ready to merge from my side.

PR #782, opened 2015-10-16
if the path is empty, return the default one

jingjingwang:


jortiz16:
thanks for finding this bug!

PR #783, opened 2015-10-24
Persist Relations

### LINKED ISSUE #779: Implement new REST API to persist relations to HDFS
We need a new REST API (with a `/persist` path prefix) to allow users to optionally persist their relations in Postgres to HDFS. Eventually we'd like to integrate this feature with MyriaL (e.g., add a `persist=true` option to the `STORE()` statement), but for now we should just expose it as a REST API (like we currently do with `DELETE`).

Details: we'll make `TupleWriter` JSON-encodable and introduce a new encodable entity `DataSink` (or an equivalent name if this causes confusion with existing nomenclature) analogous to `DataSource`, which encapsulates knowledge of how to access a particular output location (filesystem path, HDFS URL, etc.). The `TupleWriter` subclasses will initialize themselves with the `OutputStream` created by the `DataSink`, just like `FileScan` initializes itself with the `InputStream` created by a `DataSource`. Each `TupleWriter` subclass corresponds to an output format, and each `DataSink` corresponds to a location type, so we can express all possible combinations. After initializing the encoded `TupleWriter`, we can hand it to the constructor of the existing `DataOutput` operator, and we have an output leaf operator.

The `PERSIST` API will construct a query plan consisting of a single `DataOutput` operator per worker, with the HDFS path determined by our namespace conventions (i.e., a worker ID prefix for all partitions owned by that worker). The API will be asynchronous (returns immediately after scheduling the query), but we could have it return the query ID that it schedules for execution, which would allow higher-level code like myria-python to expose a synchronous interface that bubbles up query errors as exceptions. (We also could and probably should add this feature to the `DELETE` API.)

jortiz16:
Fixes #779 

In this PR, we add a new REST API call to persist relations from Postgres to HDFS. When a user calls ```/user/program/relation/persist/```, the master catalog will mark the relation as ```is_persistent``` and will send a query plan to all the workers. Each worker will read the relation using a DbQueryScan and will then write the partition to HDFS. Once all workers finish, the server will return the queryId. 

Some comments:

- All the subclasses from the TupleWriter class have an open() method. When a query plan is sent to a worker, the DataOutput operator calls the open() method during init() phase to create the OutputStream. 

- The UriSink only serves as a sink for HDFS. I know we mentioned earlier that we might want to add an option to write to S3, but currently this does not support that.

- For the persist call from Server.java, the HDFS address is hardcoded for our local cluster for vega@cs. Should we consider other ways to specify this address?  

Thanks!


src/edu/washington/escience/myria/HdfsTupleWriter.java
```diff
@@ -0,0 +1,64 @@
...
+import java.util.List;
+
+import org.apache.hadoop.fs.FSDataOutputStream;
+
+import edu.washington.escience.myria.storage.ReadableTable;
+
+/**
+ * HdfsTupleWriter is a {@link TupleWriter} that writes tuples to HDFS.
+ */
+public class HdfsTupleWriter implements TupleWriter {
```
senderista: Could we rename this class according to the format it writes rather than its expected destination (which has nothing to do with the implementation of this class)? Even something like "MyriaCustomPersistenceFormatTupleWriter" is fine.

src/edu/washington/escience/myria/HdfsTupleWriter.java
```diff
@@ -0,0 +1,64 @@
...
+  }
+
+  @Override
+  public void writeTuples(final ReadableTable tuples) throws IOException {
+    /* Write each row to the output stream */
+    for (int i = 0; i < tuples.numTuples(); ++i) {
+      String row = "";
+      for (int j = 0; j < tuples.numColumns(); ++j) {
+        row += tuples.getObject(j, i).toString();
+        row += j == tuples.numColumns() - 1 ? '\n' : ',';
```
senderista: How does this handle embedded commas/newlines? It might be more robust to use an existing CSV formatter to handle these issues, or to use control characters for the field/record delimiters (e.g., Hive uses Ctrl-A (`\001`) as its default field delimiter). `CSVFormat` in `CsvTupleWriter` is pretty heavily customizable (you'd need to expose the `CSVFormat` field in a public constructor).

src/edu/washington/escience/myria/CsvTupleWriter.java
```diff
@@ -58,6 +61,11 @@ private CsvTupleWriter(final OutputStream out, final CSVFormat csvFormat) throws
   }
 
   @Override
+  public void open(final OutputStream stream) throws IOException {
```
senderista: I don't understand how this method plays with all the existing public constructors, which take an `OutputStream`. How is this class being instantiated if the `OutputStream` is only initialized in `open()`, and why do we still need the `OutputStream` argument to the constructors?

src/edu/washington/escience/myria/JsonTupleWriter.java
```diff
@@ -50,6 +53,11 @@ public JsonTupleWriter(final OutputStream output) {
     this.output = new PrintWriter(new BufferedWriter(new OutputStreamWriter(output)));
```
senderista: As above, shouldn't we remove this constructor if we expect callers to supply the `OutputStream` in `open()`?

src/edu/washington/escience/myria/CsvTupleWriter.java
```diff
@@ -22,8 +22,11 @@
  */
 public class CsvTupleWriter implements TupleWriter {
 
+  /** Required for Java serialization. */
+  static final long serialVersionUID = 1L;
+
   /** The CSVWriter used to write the output. */
-  private final CSVPrinter csvPrinter;
+  private CSVPrinter csvPrinter;
```
senderista: Non-serializable fields should be marked `transient`.

src/edu/washington/escience/myria/HdfsTupleWriter.java
```diff
@@ -0,0 +1,64 @@
...
+/**
+ * HdfsTupleWriter is a {@link TupleWriter} that writes tuples to HDFS.
+ */
+public class HdfsTupleWriter implements TupleWriter {
+
+  /** Required for Java serialization. */
+  static final long serialVersionUID = 1L;
+
+  /** The {@link OutputStream} to which we write the data. **/
+  private FSDataOutputStream outStream;
```
senderista: Non-serializable fields should be marked `transient`.

src/edu/washington/escience/myria/JsonTupleWriter.java
```diff
@@ -34,10 +34,13 @@
...
 public class JsonTupleWriter implements TupleWriter {
 
+  /** Required for Java serialization. */
+  static final long serialVersionUID = 1L;
+
   /** The names of the columns, escaped for JSON. */
   private ImmutableList<String> escapedColumnNames;
   /** The {@link PrintWriter} wraps the {@link OutputStream} to which we write the data. */
-  private final PrintWriter output;
+  private PrintWriter output;
```
senderista: Non-serializable fields should be marked `transient`.

src/edu/washington/escience/myria/PostgresBinaryTupleWriter.java
```diff
@@ -19,8 +19,11 @@
  */
 public class PostgresBinaryTupleWriter implements TupleWriter {
 
+  /** Required for Java serialization. */
+  static final long serialVersionUID = 1L;
+
   /** The ByteBuffer to write the output. */
-  private final DataOutputStream buffer;
+  private DataOutputStream buffer;
```
senderista: Non-serializable fields should be marked `transient`.

src/edu/washington/escience/myria/PostgresBinaryTupleWriter.java
```diff
@@ -19,8 +19,11 @@
  */
 public class PostgresBinaryTupleWriter implements TupleWriter {
```
senderista: Why not make this the `TupleWriter` implementation for persisting Postgres tables to HDFS? It should be much faster to re-ingest them into Postgres using `COPY`, right?

src/edu/washington/escience/myria/io/UriSink.java
```diff
@@ -0,0 +1,31 @@
...
+import java.net.URI;
+import java.util.Objects;
+
+import org.apache.hadoop.conf.Configuration;
+import org.apache.hadoop.fs.FileSystem;
+import org.apache.hadoop.fs.Path;
+
+import com.fasterxml.jackson.annotation.JsonProperty;
+
+public class UriSink implements DataSink {
```
senderista: Don't you need a serialization UID?

src/edu/washington/escience/myria/parallel/Server.java
```diff
@@ -1057,7 +1060,49 @@ public DatasetStatus deleteDataset(final RelationKey relationKey) throws DbExcep
...
+    /* Mark the relation as is_persistent */
+    try {
+      catalog.markRelationPersistent(relationKey);
+    } catch (CatalogException e) {
+      throw new DbException(e);
+    }
+
+    /* Create the query plan for persist */
+    try {
+      Map<Integer, SubQueryPlan> workerPlans = new HashMap<>();
```
senderista: Nitpick: this could be an `ImmutableMap` (with a builder).

src/edu/washington/escience/myria/parallel/Server.java
```diff
@@ -1057,7 +1060,49 @@ public DatasetStatus deleteDataset(final RelationKey relationKey) throws DbExcep
...
+      for (Integer workerId : getWorkersForRelation(relationKey, null)) {
+        String partitionName =
+            String.format("hdfs://vega.cs.washington.edu:8020/partition%s/%s/%s/%s", workerId, relationKey
```
senderista: We should definitely be pulling the HDFS namenode URL from configuration. Should be trivial to add.

senderista: For readability, I'd like at least a hyphen between "partition" and the worker ID (or `workers/<workerId>`). Let's discuss in the student meeting but not bikeshed over it.

src/edu/washington/escience/myria/io/UriSink.java
```diff
@@ -0,0 +1,31 @@
...
+
+  public UriSink(@JsonProperty(value = "uri", required = true) final String uri) {
+    this.uri = URI.create(Objects.requireNonNull(uri, "Parameter uri cannot be null"));
+  }
+
+  @Override
+  public OutputStream getOutputStream() throws IOException {
+    Configuration conf = new Configuration();
+
+    FileSystem fs = FileSystem.get(uri, conf);
```
senderista: I don't think we want to have a separate HDFS `DataSink` impl, but maybe we could fail a bit earlier and more informatively for non-HDFS URIs? You could just compare `uri.getScheme()` to "hdfs" for now.

src/edu/washington/escience/myria/CsvTupleWriter.java
jortiz16: I added the open() call in cases which we wanted to open the stream from the worker side. Originally, these constructors were used in cases in which the coordinator would create the operator and open the outputStream in place (for example see getDatasetData() in DatasetResource which calls this CsvTupleWriter constructor). 

I could change this so the open() is only used in the MyriaCustomPersistenceFormatTupleWriter class. Or another alternative is to remove the public constructors (as suggested by your comment below) and keep only open() calls for all the writers.

senderista: I favor removing the old public constructors and upgrading all callers to use the new `open()` method. In general, I think we should avoid non-serializable parameters in operator-related constructors. @jingjingwang what do you think?

jortiz16: Hey @jingjingwang, I the latest commit here does what Tobin suggested where we remove the old constructors and use the new open methods. I think it looks cleaner as well. Let us know if you also agree when you get the chance.

src/edu/washington/escience/myria/CsvTupleWriter.java
```diff
@@ -22,39 +22,29 @@
...
-  }
-
-  /**
-   * @param out the {@link OutputStream} to which the data will be written.
-   * @param csvFormat the CSV format.
-   * @throws IOException if there is an IO exception
-   */
-  private CsvTupleWriter(final OutputStream out, final CSVFormat csvFormat) throws IOException {
-    csvPrinter = new CSVPrinter(new BufferedWriter(new OutputStreamWriter(out)), csvFormat);
+  public void open(final char separator, final OutputStream out) throws IOException {
```
senderista: The previous `open()` method should delegate to this one (I would put `separator` last for consistency).

src/edu/washington/escience/myria/TupleWriter.java
```diff
@@ -20,7 +22,15 @@
  * 
  * 
  */
-public interface TupleWriter {
+public interface TupleWriter extends Serializable {
+
+  /** Required for Java serialization. */
+  static final long serialVersionUID = 1L;
```
senderista: I don't believe this will work--see https://docs.oracle.com/javase/8/docs/api/java/io/Serializable.html: 
> such declarations apply only to the immediately declaring class--serialVersionUID fields are not useful as inherited members.

src/edu/washington/escience/myria/operator/DataOutput.java
```diff
@@ -61,6 +69,9 @@ protected void consumeTuples(final TupleBatch tuples) throws DbException {
   @Override
   protected void init(final ImmutableMap<String, Object> execEnvVars) throws DbException {
     try {
+      if (dataSink != null) {
+        tupleWriter.open(dataSink.getOutputStream());
+      }
       tupleWriter.writeColumnHeaders(getChild().getSchema().getColumnNames());
```
senderista: How can this succeed if `tupleWriter.open()` isn't called (when `dataSink` is `null`)?

src/edu/washington/escience/myria/parallel/Server.java
```diff
@@ -1057,7 +1066,51 @@ public DatasetStatus deleteDataset(final RelationKey relationKey) throws DbExcep
...
+    /* Mark the relation as is_persistent */
+    try {
+      catalog.markRelationPersistent(relationKey);
+    } catch (CatalogException e) {
+      throw new DbException(e);
+    }
 
+    /* Create the query plan for persist */
+    try {
+      ImmutableMap<Integer, SubQueryPlan> workerPlans = null;
```
senderista: I don't understand how this can work--you're overwriting `workerPlans` on each iteration of the `for` loop? You presumably want to declare a new `Builder` here and call `Builder.put()` on each iteration, then call `Builder.build()` after the loop is finished, passing the resulting `ImmutableMap` to `submitQuery()`.

src/edu/washington/escience/myria/parallel/Server.java
```diff
@@ -1057,7 +1066,51 @@ public DatasetStatus deleteDataset(final RelationKey relationKey) throws DbExcep
...
+      for (Integer workerId : getWorkersForRelation(relationKey, null)) {
+        String partitionName =
+            String.format(persistURI + "/partition-%s/%s/%s/%s", workerId, relationKey.getUserName(), relationKey
```
senderista: Not sure what I suggested before, but it might be better to have a "reserved" prefix as a separate path component to denote this namespace, maybe something like `myria-system`? I don't have any real preferences here--maybe we could discuss for a couple minutes in the next Myria meeting and get consensus.

src/edu/washington/escience/myria/operator/DataOutput.java
jortiz16: The reason why I added this was because the DataOutput operator works for two cases:

Case #1: The server outputs data from the workers (for example in the case where someone downloads data) 

Case #2: The case where each worker writes the data to hdfs. 

I added the null case here because for Case #1, dataSink is null. We do not have a dataSink for downloading data.  To answer your previous question, this would still work because I would open the TupleWriter instance in DatasetResource.java before sending it to the query plan with the DataOutput operator. The problem with this is that there is no way to ensure that the TupleWriter is open (if dataSink is null) as you've mentioned.

The other way to do this is to get rid of this if-statement and require the tupleWriter to only be opened at the DataOutput operator itself. This would require me to create another type of valid dataSink for the "downloading data" case. I'm currently working on this version since I think it will be cleaner at the end. What do you think?

senderista: I completely agree that a new `DataSink` impl for downloading data is the way to go!

jortiz16:
The main change now is the addition of PipeSink (I couldn't think of a better name). If we're using a DataOutput operator, we now require a DataSink. This mostly required changes in Server.java for methods that would download data and for methods that used the profiling tool. 

In this commit I also changed the code based on your previous suggestions (ImmutableMap fix, the myria-system namespace for hdfs, and fixed the constructors for CsvTupleWriter)

jortiz16:
For the DataOutput constructor with the ```null``` dataSink, I ended up getting rid of it. Now DataOutput always requires a valid dataSink. The only method  that was using the null dataSink call was the getResourceLog() method in Server.java (which should now be fixed).

senderista:
LGTM

Issue #781, opened 2015-10-07
In setup_cluster.py, `clean-catalog` flag should be implied when the master catalog doesn't already exist


BrandonHaynes:
This confused me.

jortiz16:
plus one! I was very confused

PR #794, opened 2016-01-22
always create a clean catalog when there is no existing one

jingjingwang:
Fixes https://github.com/uwescience/myria/issues/781.


BrandonHaynes:
LGTM -- at some point the ad hoc SSH command string construction should be centralized and the parameters properly encoded.

PR #796, opened 2016-01-28
Add support for loading Tipsy data from HDFS

BrandonHaynes:
There are several classes that all contain their own "load from x" functionality; this should be centralized at some point.

### DIFF
--- a/src/edu/washington/escience/myria/operator/TipsyFileScan.java
+++ b/src/edu/washington/escience/myria/operator/TipsyFileScan.java
@@ -332,12 +332,16 @@ public class TipsyFileScan extends LeafOperator {
       } else if (uri.getScheme().equals("hdfs")) {
         return openHdfsInputStream(uri);
       } else {
-        return openFileInputStream(filenameOrUrl);
+        return uri.toURL().openStream();
       }
     } catch (IllegalArgumentException e) {
       return openFileInputStream(filenameOrUrl);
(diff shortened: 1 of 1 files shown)


src/edu/washington/escience/myria/operator/TipsyFileScan.java
```diff
@@ -322,7 +326,13 @@ protected Schema generateSchema() {
 
   private static InputStream openFileOrUrlInputStream(String filenameOrUrl) throws DbException {
     try {
-      return new URI(filenameOrUrl).toURL().openConnection().getInputStream();
+      URI uri = new URI(filenameOrUrl);
+      if(uri.getScheme() == null)
```
senderista: spaces around if/catch (we really should have checkstyle running in travis rather than manually reviewing for style errors).

src/edu/washington/escience/myria/operator/TipsyFileScan.java
```diff
@@ -341,4 +351,14 @@ private static InputStream openFileInputStream(String filename) throws DbExcepti
       throw new DbException(e);
     }
   }
+
+  private static InputStream openHDFSInputStream(final URI uri) throws DbException {
```
senderista: java convention would be "Hdfs"

senderista:
@BrandonHaynes Can we make these trivial changes and merge into `master`?

PR #806, opened 2016-02-12
Update README.md

domoritz:
Instead of using a symlink to a markdown file in the docs, use a separate readme.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,1 +1,5 @@
-docs/index.markdown
+myria
+=====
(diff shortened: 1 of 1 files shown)


jingjingwang:
Does this mean that we expect the user to go to the myria website instead of reading the local README.md? I think at some point our consensus was having an unique readme file for each repo that can be accessed either by checking out a repo or being pulled to the website.

domoritz:
The readme for the repository shouldn't have anything to do with the docs that are shown on the website. The consensus was to pull the docs from each repo so that we can maintain them in the context of the code. The docs are not the readme, though.

senderista:
@jingjingwang could you comment on status of this PR? Didn't we decide on a different approach (pulling the README docs for each repo directly into the website repo)?

Issue #797, opened 2016-02-04
Create generic S3 protocol handler in `UriSource` and `UriSink`


senderista:
We should have a generic S3 protocol handler for import/export, using the Hadoop S3 `FileSystem` implementation (the [`s3a`](http://hadoop.apache.org/docs/r2.7.2/hadoop-aws/tools/hadoop-aws/index.html) one to be precise, since it's built on the AWS SDK, and hence among other things allows us to handle credentials securely). This is necessary for exporting any data to S3, and for importing any S3 data that isn't publicly readable (and thus accessible through HTTP). Specific optimizations for parallel S3 import/export will be addressed in separate issues.

This will require us to obtain credentials for calling into the AWS SDK. The most secure way to handle credentials is to require them to be present on an EC2 instance via an IAM role used to launch the instance: http://docs.aws.amazon.com/AWSEC2/latest/UserGuide/iam-roles-for-amazon-ec2.html. This would require us to add support for IAM roles to the deployment script, which shouldn't be difficult, but would also require the AWS identity running the deployment script to have the `PassRole` permission for the IAM role they want to pass to their cluster. Alternatively, we could propagate the user's AWS credentials directly to the cluster by getting them from `boto`, passing them to our Ansible playbook, and writing them as files on the remote machine for use by [`envdir`](http://thedjbway.b0llix.net/daemontools/envdir.html), then wrapping all launch commands for Myria executables in an `envdir` invocation with the location of the credentials. This isn't as secure as launching under an IAM role, but should be secure enough, provided we create a dedicated unprivileged user for the `myria` service, with exclusive read permissions on the `envdir` files containing AWS credentials. I favor the first approach (launching the cluster under an IAM role), since it's much simpler and more secure, but we have to evaluate the inconvenience to users who may have difficulty obtaining `PassRole` privileges. In any case, I think the first approach should be implemented first.

It [appears](https://issues.apache.org/jira/browse/HADOOP-11571) that the `s3a` implementation was significantly improved in Hadoop 2.7, so we should probably upgrade to the latest 2.7 version before deploying this feature (unless it's incompatible with REEF, which is [still](https://github.com/markusweimer/incubator-reef/blob/master/pom.xml#L51) at Hadoop 2.6.0).

senderista:
@jingjingwang would appreciate your feedback.

BrandonHaynes:
[...]

PR #811, opened 2016-02-20
Adding S3 to export/import

jortiz16:
Addresses #797, including ByteSink

I ran into a bit of trouble with the UploadDownloadS3Test. When I uploaded the data to S3, I did not have the ability to exclude the column headers via the CsvTupleWriter constructor (actually, the DataOutput operator is the one responsible for writing the headers). So when I tried to download the data, I couldn't use the JsonUtils.ingest method. I had to write the whole plan (I primarily needed the FileScan operator to handle skipping the first line for me). We need to make sure that the import/export call that will be built in Raco can run the whole upload/download roundtrip correctly.

Also, I tried using the sortColumns parameter for the DbScan operator to have deterministic results, but really my main problem was during the download call (once the data from S3 was ingested -- line 89). During download, if both workers had tuples, it is not certain which worker will provide their tuples first to the result. So I ended up adding another file to help me hash all the tuples to 1 worker instead. 

Thanks! And as always, I'm open to other suggestions.


src/edu/washington/escience/myria/io/ByteSink.java
```diff
@@ -0,0 +1,24 @@
...
+
+import java.io.ByteArrayOutputStream;
+import java.io.IOException;
+import java.io.OutputStream;
+
+public class ByteSink implements DataSink {
+  /** Required for Java serialization. */
+  private static final long serialVersionUID = 1L;
+
+  final ByteArrayOutputStream writerOutput;
```
senderista: If we materialize this in the constructor (which I don't think we should do), then it should be marked `transient`.

src/edu/washington/escience/myria/io/ByteSink.java
```diff
@@ -0,0 +1,24 @@
...
+
+  public ByteSink() {
+    writerOutput = new ByteArrayOutputStream();
```
senderista: Is there a good reason not to materialize the `OutputStream` in `getOutputStream()`, just like all the `DataSource` implementations?

src/edu/washington/escience/myria/io/UriSource.java
```diff
@@ -44,19 +44,26 @@
    * {@link InputStream}.
    *
    * @param uri the Uniform Resource Indicator (URI) of the data source.
+   * @throws URISyntaxException
    */
   @JsonCreator
-  public UriSource(@JsonProperty(value = "uri", required = true) final String uri) {
-    this.uri = Objects.requireNonNull(uri, "Parameter uri to UriSource may not be null");
+  public UriSource(@JsonProperty(value = "uri", required = true) final String uri) throws URISyntaxException {
+    URI parsedUri = URI.create(Objects.requireNonNull(uri, "Parameter uri to UriSource may not be null"));
```
senderista: Rather than round-tripping this parsed URI, we could introduce a `transient final` instance variable `parsedUri` which would be set by the constructor and used by `getInputStream()`.

systemtest/edu/washington/escience/myria/systemtest/UploadDownloadS3Test.java
```diff
@@ -0,0 +1,98 @@
...
+    String filePath = Paths.get("testdata", "filescan", "simple_two_col_int_to_hash.txt").toString();
+    DataSource relationSource = new FileSource(filePath);
+    RelationKey relationKeyUpload = RelationKey.of("public", "adhoc", "upload");
+    Schema relationSchema = Schema.ofFields("x", Type.INT_TYPE, "y", Type.INT_TYPE);
+
+    JsonAPIUtils.ingestData("localhost", masterDaemonPort, ingest(relationKeyUpload, relationSchema, relationSource,
+        ' ', new RoundRobinPartitionFunction(workerIDs.length)));
+
+    /* File to upload and download */
+    String fileName = String.format("s3://myria-test/test.txt");
```
senderista: We need to randomize this location with a timestamp or UUID to allow tests to be run concurrently.

systemtest/edu/washington/escience/myria/systemtest/UploadDownloadS3Test.java
```diff
@@ -0,0 +1,98 @@
...
+
+    /* Construct the query and upload data */
+    ExchangePairID serverReceiveID = ExchangePairID.newID();
+    DbQueryScan workerScan = new DbQueryScan(relationKeyUpload, relationSchema);
+    CollectProducer workerProduce = new CollectProducer(workerScan, serverReceiveID, MASTER_ID);
```
senderista: `workerProducer`?

systemtest/edu/washington/escience/myria/systemtest/UploadDownloadS3Test.java
```diff
@@ -0,0 +1,98 @@
...
+
+    HashMap<Integer, RootOperator[]> workerPlans = new HashMap<Integer, RootOperator[]>();
```
senderista: `Map`

systemtest/edu/washington/escience/myria/systemtest/UploadDownloadS3Test.java
```diff
@@ -0,0 +1,98 @@
...
+    for (int workerID : workerIDs) {
+      workerPlans.put(workerID, new RootOperator[] { workerProduce });
+    }
+    CollectConsumer serverCollect = new CollectConsumer(relationSchema, serverReceiveID, workerIDs);
```
senderista: Consistency: `workerProducer`/`serverConsumer` or `workerCollect`/`serverCollect`?

systemtest/edu/washington/escience/myria/systemtest/UploadDownloadS3Test.java
```diff
@@ -0,0 +1,98 @@
...
+    DataOutput masterRoot = new DataOutput(sortOperator, new CsvTupleWriter(), dataSink);
+    server.submitQueryPlan(masterRoot, workerPlans).get();
+
+    /* Read the data back in from S3 and shuffle to one worker */
+    RelationKey relationKeyDownload = RelationKey.of("public", "adhoc", "download");
+    DataSource relationSourceS3 = new UriSource(fileName);
+
+    ExchangePairID workerReceiveID = ExchangePairID.newID();
+    FileScan serverFileScan = new FileScan(relationSourceS3, relationSchema, ',', null, null, 1);
+    GenericShuffleProducer serverProduce =
```
senderista: `serverProducer`

systemtest/edu/washington/escience/myria/systemtest/UploadDownloadS3Test.java
```diff
@@ -0,0 +1,98 @@
...
+        new GenericShuffleProducer(serverFileScan, workerReceiveID, workerIDs, new SingleFieldHashPartitionFunction(
+            workerIDs.length, 0));
+    GenericShuffleConsumer workerConsumer =
+        new GenericShuffleConsumer(relationSchema, workerReceiveID, new int[] { MASTER_ID });
+    DbInsert workerInsert = new DbInsert(workerConsumer, relationKeyDownload, true);
+    HashMap<Integer, RootOperator[]> workerPlansInsert = new HashMap<Integer, RootOperator[]>();
```
senderista: `Map`

systemtest/edu/washington/escience/myria/systemtest/UploadDownloadS3Test.java
```diff
@@ -0,0 +1,98 @@
...
+      workerPlansInsert.put(workerID, new RootOperator[] { workerInsert });
+    }
+    server.submitQueryPlan(serverProduce, workerPlansInsert).get();
+
+    String dstData =
+        JsonAPIUtils.download("localhost", masterDaemonPort, relationKeyDownload.getUserName(), relationKeyDownload
+            .getProgramName(), relationKeyDownload.getRelationName(), "json");
+
+    String srcData =
+        "[{\"x\":1,\"y\":2},{\"x\":1,\"y\":2},{\"x\":1,\"y\":4},{\"x\":1,\"y\":4},{\"x\":1,\"y\":6},{\"x\":1,\"y\":6}]";
```
senderista: Not a big fan of storing the JSON form of the test data independently of its source file, but if constructing the JSON string in the right order from the source data is ridiculously complicated then I guess it's OK.

senderista:
Tests look good (mod a few nitpicks)!

PR #818, opened 2016-03-22
exclude old jersey-server from our repo

jingjingwang:
otherwise we have both 1.9 and 2.14 jersey-server and cluster can't launch.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -107,7 +107,7 @@ dependencies {
     exclude group: "com.sun.jersey"
   }
 
-  compile ("org.apache.hadoop:hadoop-common:${hadoopVersion}") {
+  compile ("org.apache.hadoop:hadoop-client:${hadoopVersion}") {
     /* Hadoop depends on Jersey 1 but we don't need it. Exclude it to prevent
      * Classloader picking the wrong version of Jersey classes. */
     exclude group: "com.sun.jersey"
--- a/build.gradle
+++ b/build.gradle
@@ -101,13 +101,17 @@ ext.jmxetricsVersion = "1.0.8"
 ext.hadoopVersion = "2.7.2"
 
 dependencies {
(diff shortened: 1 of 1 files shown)


jingjingwang:
This may explain the issue that @parmitam mentioned this morning. I think Travis passed because it happened to pick the correct version.

build.gradle
```diff
@@ -101,13 +101,17 @@ ext.jmxetricsVersion = "1.0.8"
...
 
 dependencies {
-  compile ("org.apache.hadoop:hadoop-client:${hadoopVersion}") {
+
+  /* Note this jar is fetched from our private Maven repo (patched for HADOOP-12807) */
+  compile ("org.apache.hadoop:hadoop-aws:${hadoopVersion}") {
+    exclude group: "com.sun.jersey"
+  }
+
+  compile ("org.apache.hadoop:hadoop-common:${hadoopVersion}") {
```
senderista: why the change from `hadoop-client` to `hadoop-common`?

jingjingwang: Because `haoop-aws` also depends on `hadoop-common` which depends on `com.sun.jersey`. Putting `hadoop-client` there is not enough to kick 1.9 away.

jingjingwang: PS `./gradlew dependencies` has the whole dependency tree.

senderista: I'm actually not sure how this compiles and runs tests without the `hadoop-client` dependency, since it doesn't appear that `hadoop-common` pulls in the YARN and HDFS client libs we need. In any case,  wouldn't it be better to leave the explicit dependency on `hadoop-client` and `hadoop-aws` (since those are the libraries we actually consume), and exclude `com.sun.jersey` just from `hadoop-common`?

jingjingwang:
I'm not sure if I fully understood your comment but I'm keeping the exclude from both `hadoop-aws` and `hadoop-client` explicitly.

Issue #817, opened 2016-03-16
OutOfMemory does not lead to query failure immediately


jingjingwang:
I have observed such a case where I saw OutOfMemory in worker's stderr, but the query was still running for an infinite amount of time until I had to kill it. It happens sometimes when the heap size limit is close to the working set size. We need to investigate it more. 
I remember that we observed similar cases before -- having a query running forever. It may explain them.

senderista:
Is this on the `reef` branch?

senderista:
To elaborate, on the `reef` branch, if a worker exceeds allocated memory, then the YARN NodeManager will kill it. Or, if the machine is low on memory, then the Linux OOM killer might terminate it. Either way, the `Driver` will time out on heartbeats from that worker and declare the `Evaluator` dead, which will fire a `LostHeartbeatException` from the coordinator, which should always fail the query.

jingjingwang:
No it's the main branch. I'll look at it more deeply.

jingjingwang:
The PR is merged and I believe this issue is resolved.

Issue #615, opened 2014-09-11
OOM causes query to hang


dhalperi:
71/72 workers have finished. I checked the one that failed and saw this message:

```
Exception: java.lang.OutOfMemoryError thrown from the UncaughtExceptionHandler in thread "IPC worker#75"

Exception: java.lang.OutOfMemoryError thrown from the UncaughtExceptionHandler in thread "Nonblocking query executor#11"

Exception: java.lang.OutOfMemoryError thrown from the UncaughtExceptionHandler in thread "IPC worker#60"

Exception: java.lang.OutOfMemoryError thrown from the UncaughtExceptionHandler in thread "IPC worker#41"

Exception: java.lang.OutOfMemoryError thrown from the UncaughtExceptionHandler in thread "New I/O worker #54"

Exception: java.lang.OutOfMemoryError thrown from the UncaughtExceptionHandler in thread "IPC connection pool global timer#0"

Exception: java.lang.OutOfMemoryError thrown from the UncaughtExceptionHandler in thread "IPC worker#28"

Exception: java.lang.OutOfMemoryError thrown from the UncaughtExceptionHandler in thread "New I/O worker #43"

Exception: java.lang.OutOfMemoryError thrown from the UncaughtExceptionHandler in thread "New I/O worker #5"
ERROR 2014-09-11 13:49:52,489 [New I/O worker #1] Worker$1 - Uncaught exception in thread: Thread[New I/O worker #1,5,MyriaWorkerThreadGroup]
java.lang.OutOfMemoryError: Java heap space
WARN  2014-09-11 13:49:52,490 [New I/O worker #35] IPCMessageHandler - Unknown error occur in unmanaged Netty Channel: [Client,#? => #?] [id: 0x699fa1de, 0.
0.0.0/0.0.0.0:32906 :> vega.cs.washington.edu/128.208.3.110:1777], close directly.
java.lang.OutOfMemoryError: Java heap space
ERROR 2014-09-11 13:49:52,490 [New I/O worker #85] Worker$1 - Uncaught exception in thread: Thread[New I/O worker #85,5,MyriaWorkerThreadGroup]
java.lang.OutOfMemoryError: Java heap space
ERROR 2014-09-11 13:49:52,490 [IPC worker#77] Worker$1 - Uncaught exception in thread: Thread[IPC worker#77,5,MyriaWorkerThreadGroup]
java.lang.OutOfMemoryError: Java heap space
[...]

PR #819, opened 2016-03-23
make JVM exit when OOM

jingjingwang:
When the JVM is close to an OOM state, the OOM error can be thrown at any time by any thread, even threads that are not used for query processing. The old implementation takes a long way to propagate an OOM error from the lowest level to Worker level, and the system state can crash during this process so that the query is still marked as running but the JVM cannot respond anymore. On the other hand, it is unnecessary to clean up after an OOM since the JVM is in an OOM state anyway. So we terminate the worker process when an OOM is detected.


jingjingwang:
Although it's a brute force solution, since we're switching to one-query-per-worker, I believe it's still acceptable. I'm not sure how it should be merged with the reef branch though.

jingjingwang:
Solves https://github.com/uwescience/myria/issues/615 and https://github.com/uwescience/myria/issues/817.

senderista:
I don't see any reason to merge this into the `reef` branch.

PR #823, opened 2016-04-24
Support sinking to local file system

BrandonHaynes:
Sink for pushing directly to the file system, and an integration test that verifies that the export is successfully written.


senderista:
You also need to add a new JSON subtype annotation to `edu.washington.escience.myria.io.DataSink`.

PR #829, opened 2016-05-16
Add code formatter

jingjingwang:


jingjingwang:
The last commit will change once we have the pending PRs being merged, but the first two is the key so I think maybe it's okay for you to take a look even now.

build.gradle
```diff
@@ -69,8 +78,26 @@ check.dependsOn testsuitesClasses
 /* Build the Eclipse .classpath file first, because if Gradle's build fails
  * we'll want to use Eclipse to debug it. */
 test.dependsOn eclipseClasspath
+/* Test if all *.java have been formatted. Run ./gradlew format to format all */
```
senderista: Shouldn't this comment move down to line 99?

senderista:
LGTM

jingjingwang:
Nice catch! (It used to be there)

Issue #816, opened 2016-03-08
Implement `DbQueryExecute` operator


senderista:
We need a generic operator for passing SQL to Postgres as-is. We can initially restrict MyriaL access to this operator to just creating views and indexes, but it would be useful to have the capability to execute arbitrary SQL on a given list of workers for user-generated query plans.

For particular applications like `CREATE VIEW` and `CREATE INDEX`, we can track the in-progress status of the created entities in the catalog and guarantee idempotence (and atomicity WRT visibility from the catalog APIs). For executing generic DDL, we need to use some distributed transaction facility like `PREPARE TRANSACTION` (http://www.postgresql.org/docs/9.4/static/sql-prepare-transaction.html). The coordinator will need to provide these distributed guarantees (possibly in a REST API implementation).

jortiz16:
@senderista what do you mean by keeping progress of the entities through the catalog? Do you mean adding an extra column to the relations table (sort of like what we did for the delete feature)?

I need this feature for the PSLA integration with Myria. I could help implement implement the DbExecuteOperator for at least the ```create view``` and ```create index``` cases. How does that sound?

senderista:
Yes, I believe I meant recording the status of the operation on each partition. Idempotent retry of a "create view/index" operation would just create the view/index/whatever on each partition where it doesn't exist (similar to the delete operation).

What do you need exactly for the PSLA integration? Do you just need views/indexes? I'd rather avoid implementing a generic passthrough operator if we don't need it (since it would require distributed transactions).

jortiz16:
Yes, I need it in order to create views. I mean, I could do it the old-fashioned way -- (creating .sql scripts, broadcasting to all the workers, and then running each script locally). Since both ```create index``` and ```create views``` is idempotent, I don't see harm in supporting just these two operators. 

I don't need the "generic" (run anything on postgres) operation. This is not only more complicated but can be harmful too. Someone could run "DROP tables" or really mess up the state of the underlying database.

senderista:
OK, so you would implement a new operator for each of `CREATE VIEW` and `CREATE INDEX`? That should be fine. Looks like there's already index support in `edu.washington.escience.myria.accessmethod.JdbcAccessMethod`. Can you just use that? Could you add view support to the same class (with Postgres-specific extensions being delegated to helper methods)? Also, do you need materialized views?

BrandonHaynes:
[...]

PR #830, opened 2016-05-17
DbExecute Operators

jortiz16:
We want to support the ability to do the following: create indexes in Postgres and also run arbitrary Postgres UDFs. This PR only supports Postgres UDFs, but @parmitam is planning to extend this in the near future to support Python UDFs. 

There might be a few things up for discussion, but this PR includes the ability to create views, indexes, and UDFs through provided encodings. The REST API can read these encodings and then make a call to the server (which creates the query plan with the corresponding operators and adds metadata to the Master catalog). 

I'm open to other suggestions/renaming for the encodings. Also, another thing to notice is that I did not record the creation of views in the master catalog because this is not an operation we will want to expose to users at the MyriaL level.


src/edu/washington/escience/myria/operator/DbCreateView.java
```diff
@@ -0,0 +1,62 @@
...
+  protected void init(final ImmutableMap<String, Object> execEnvVars) throws DbException {
+    /* Retrieve connection information from the environment variables, if not already set */
+    if (connectionInfo == null && execEnvVars != null) {
+      connectionInfo = (ConnectionInfo) execEnvVars.get(MyriaConstants.EXEC_ENV_VAR_DATABASE_CONN_INFO);
+    }
+    /* Open the database connection */
+    accessMethod = AccessMethod.of(connectionInfo.getDbms(), connectionInfo, false);
+
+    /* Add the indexes to the relation */
+    accessMethod.createViewIfNotExists(viewName, viewQuery);
```
senderista: I thought we had agreed that to preserve idempotence the local operator would always overwrite any existing view definition?

src/edu/washington/escience/myria/operator/DbCreateUDF.java
```diff
@@ -0,0 +1,75 @@
...
+  protected void init(final ImmutableMap<String, Object> execEnvVars) throws DbException {
+    /* Retrieve connection information from the environment variables, if not already set */
+    if (connectionInfo == null && execEnvVars != null) {
+      connectionInfo = (ConnectionInfo) execEnvVars.get(MyriaConstants.EXEC_ENV_VAR_DATABASE_CONN_INFO);
+    }
+    /* Open the database connection */
+    accessMethod = AccessMethod.of(connectionInfo.getDbms(), connectionInfo, false);
+
+    /* Drop the table */
+    accessMethod.runCommand(udfDefinition);
```
senderista: What is the purpose of defining a separate operator for UDF creation when it simply executes an arbitrary user-provided SQL statement (with no validation!)? Why not just have the `createUDF` API call a generic `DbExecute` operator?

src/edu/washington/escience/myria/api/encoding/CreateIndexEncoding.java
```diff
@@ -0,0 +1,22 @@
...
+import edu.washington.escience.myria.accessmethod.AccessMethod.IndexRef;
+
+/**
+ * 
+ */
+public class CreateIndexEncoding extends MyriaApiEncoding {
+  @Required
+  public RelationKey relationKey;
+  @Required
+  public Schema schema;
```
senderista: I question the necessity of including the schema. Isn't this redundant information since, given `relationKey`, we can look up the schema from the catalog, and infer the columns that should be indexed from the offsets in `indexes`?

src/edu/washington/escience/myria/coordinator/MasterCatalog.java
```diff
@@ -146,6 +148,12 @@
       "UPDATE queries "
     + "SET status = '" + QueryStatusEncoding.Status.UNKNOWN.toString() + "'"
     + "WHERE status = '" + QueryStatusEncoding.Status.ACCEPTED.toString() + "';";
+  private static final String CREATE_REGISTERED_UDFS =
+      "CREATE TABLE registered_udfs (\n"
+          + "    udf_id INTEGER NOT NULL, \n"
+          + "    udf_name INTEGER NOT NULL, \n"
```
senderista: Shouldn't this be `TEXT`?

src/edu/washington/escience/myria/operator/DbCreateUDF.java
```diff
@@ -0,0 +1,75 @@
...
+  @Override
+  protected void init(final ImmutableMap<String, Object> execEnvVars) throws DbException {
+    /* Retrieve connection information from the environment variables, if not already set */
+    if (connectionInfo == null && execEnvVars != null) {
+      connectionInfo = (ConnectionInfo) execEnvVars.get(MyriaConstants.EXEC_ENV_VAR_DATABASE_CONN_INFO);
+    }
+    /* Open the database connection */
+    accessMethod = AccessMethod.of(connectionInfo.getDbms(), connectionInfo, false);
+
+    /* Drop the table */
```
senderista: change doc comment

src/edu/washington/escience/myria/coordinator/MasterCatalog.java
```diff
@@ -1494,6 +1504,88 @@ protected Void job(final SQLiteConnection sqliteConnection) throws CatalogExcept
...
+      throw new CatalogException("Catalog is closed.");
+    }
+
+    /* Do the work */
+    try {
+      queue.execute(new SQLiteJob<Void>() {
+        @Override
+        protected Void job(final SQLiteConnection sqliteConnection) throws CatalogException, SQLiteException {
+          try {
+            long udf_id = sqliteConnection.getLastInsertId();
```
senderista: This seems dicey, even if it's safe given that only one thread is using this connection. I thought our general approach to primary keys (admittedly contradicting general performance-oriented practice) was to avoid unnecessarily synthesizing primary keys (e.g. integer IDs) if we already have a suitable (possibly composite) primary key. Isn't `udf_name` already such a key?

senderista:
fixes #816

src/edu/washington/escience/myria/operator/DbCreateUDF.java
jortiz16: By "generic" ```DbExecute``` operator do you mean having another new operator exclusively handle UDF creation? 

We can handle it like we did with DbCreateView/DbCreateIndex where we asked for specific input parameters. 

Or we could just add some sort of validation before calling "runCommand". If we did validate, what would we check for? Some things I could think of include: 1) checking for "CREATE FUNCTION" keyword 2) checking if there is no additional command running after the UDF creation command

senderista: I'm thinking that the `createUDF` API implementation can perform any validation and hand off execution to a generic SQL execution operator which itself does no validation. There doesn't need to be anything UDF-specific in the operator itself (and there isn't right now). As long as we don't create an encoding for this operator, it can only be instantiated programmatically, so it doesn't seem that dangerous to me. Maybe the validation at the API layer could include checking for `CREATE FUNCTION` at the beginning of the SQL string and `LANGUAGE SQL;` at the end (not sure if the latter would be too restrictive)?

jortiz16:
Still testing, not yet ready for review. I only pushed since @parmitam needed to see changes. Thanks!

src/edu/washington/escience/myria/coordinator/MasterCatalog.java
```diff
@@ -113,6 +114,7 @@
     + "    col_index INTEGER NOT NULL,\n"
     + "    col_name TEXT,\n"
     + "    col_type TEXT NOT NULL,\n"
+    + "    is_indexed INTEGER NOT NULL, \n"
```
jortiz16: So I basically relied on createIndexes method that already existed in the JdbcAccessMethod class. These methods did not have the ability to specify index order. Should I add this feature to the createIndex methods and let users specify sorting order through the createIndex encoding? I don't think this would be difficult to do....Although I'm not yet sure how users should specify this order

jortiz16: The easiest thing I could think of would be to let users specify the CreateIndex encoding as the following:

{
  "relationKey": {
    "userName": "jwang",
    "programName": "global_join",
    "relationName": "smallTable"
  },
  "schema": {
    "columnTypes": [
      "LONG_TYPE",
      "LONG_TYPE"
    ],
    "columnNames": [
      "col1",
      "col2"
    ]
  },
  "indexes": [
    0,1
  ],
  "indexIsAscending": [0,1]
}

Perhaps "indexIsAscending" isn't the best name, but ultimately it would create an index like the following:

"CREATE INDEX indexName ON "jwang:global_join:smallTable" (col1 DESC, col2 ASC)"

jortiz16: Sorry, should be "columnsAreAscending" not indexes.

jortiz16: Wow, I totally missed the fact that IndexRef type already holds information for the index order.  Nevermind!

senderista: I guess I should take a step back and ask why users would want to specify
index ordering. Normally it would be to optimize the final ordering of
results, but since we don't support ORDER BY anyway, this isn't (currently)
relevant. Even if we did support ORDER BY, an extra sort operator at the
root probably isn't important enough for a user to try to optimize away
with clever indexes. So I'm not sure exactly what I was thinking when I
proposed this. Can you think of a compelling use case for user-specified
index ordering? Would it help us do well on any benchmarks?

senderista: So `IndexRef ` contains information on the index ordering that can be used by your new operator, but I don't see where that's persisted in the catalog metadata. And the only existing implementation `createIndexIfNotExistPostgres` apparently ignores that information (but I guess that's something you can fix)!

jortiz16: I can't think of specific cases where this might be useful. I forgot about lack of support for ORDER BY. In the past, creating indexes was useful when pushing down selection operations to Postgres. For example, with the cosmo data, I would need to find particles that belonged to a particular galaxy (grpId). For that scenario, it was useful to have an index for the grpID column, but if I were to run that today, I would also need enable  pushing down the computation to postgres. 

Long story short, I don't think there is a strong case for index ordering. It seems like indexes will be useful only for selections that are pushed down to postgres....

jortiz16: ```createIndexIfNotExistPostgres``` doesn't ignore the order from the ```indexRef``` list. The ```getIndexColumns``` in that method creates the sql string with the column names along with the order. The default is "ASC" if no order is specified.

senderista: Got it, thanks.

senderista: They could also be quite useful for joins and aggregates that are pushed
into Postgres...

jortiz16:
Ready for another review.

Changes:
1) Changed any mention of "UDF" to "Function"
2) Made sure all operations are idempotent. Views and Functions are internally called through a ```CREATE OR REPLACE``` call. Not yet sure if this information should somehow be exposed to the user.  
3) For indexes, we now add order metadata to the catalog
4) Functions are validated and passed through a generic ```dbExecute``` operator. There is no longer a ```DbCreateFunction``` operator. 
5) For the ```registered_functions``` table in the catalog, ```function_name``` is now the primary key

src/edu/washington/escience/myria/api/encoding/CreateFunctionEncoding.java
```diff
@@ -5,11 +5,14 @@
 
 import java.util.Set;
 
+import edu.washington.escience.myria.Schema;
+
 /**
  *
  */
 public class CreateFunctionEncoding extends MyriaApiEncoding {
   @Required public String functionName;
```
senderista: A bit of a nitpick: is it necessary for all these fields to have the prefix `function`? This isn't a pattern we follow elsewhere, and unless it disambiguates something, I feel it just adds noise.

src/edu/washington/escience/myria/coordinator/MasterCatalog.java
```diff
@@ -115,7 +115,7 @@
           + "    col_name TEXT,\n"
           + "    col_type TEXT NOT NULL,\n"
           + "    is_indexed INTEGER NOT NULL, \n"
-          + "    ascending_order BOOLEAN,\n"
+          + "    ascending_order INTEGER, \n"
```
senderista: What's the meaning of this integer field? Since it's a nullable attribute I assume that it's equivalent to a nullable Boolean? If that is the case, then maybe a name that suggested boolean-ness would be better, like `is_ascending_index` or `has_ascending_index_order`? Otherwise I'm left wondering if it just represents some arbitrary encoding of index order into integers (like `+/-1`, say, rather than `0/1`).

src/edu/washington/escience/myria/coordinator/MasterCatalog.java
```diff
@@ -1711,10 +1713,13 @@ protected Void job(final SQLiteConnection sqliteConnection)
    */
 
   public void registerFunction(
-      @Nonnull final String functionName, @Nonnull final String functionDefinition)
+      @Nonnull final String functionName,
```
senderista: These parameter names should change, of course, if the encoding field names change.

src/edu/washington/escience/myria/coordinator/MasterCatalog.java
```diff
@@ -1728,16 +1733,17 @@ public void registerFunction(
                 protected Void job(final SQLiteConnection sqliteConnection)
                     throws CatalogException, SQLiteException {
                   try {
-                    deleteFunctionIfExists(sqliteConnection, functionName, false);
+                    deleteFunctionIfExists(sqliteConnection, functionName);
```
senderista: Is this `DELETE` necessary if we have overwrite semantics? Since the `INSERT` will always be done in a single transaction, I don't see why it's necessary if we can overwrite.

jortiz16: You mean I should replace with an ```UPDATE``` instead? If we don't delete the row I thought we would run into issues if we try to ```INSERT``` with the same key again

src/edu/washington/escience/myria/parallel/Server.java
```diff
@@ -1099,52 +1101,63 @@ public long createView(
   /**
    * Create a function and register it in the catalog
    */
-  public long createFunction(
-      final String functionName, final String functionDefinition, final Set<Integer> workers)
+  public String createFunction(
+      final String functionName,
```
senderista: These parameter names should change, of course, if the encoding field names change.

src/edu/washington/escience/myria/parallel/Server.java
```diff
@@ -1099,52 +1101,63 @@ public long createView(
...
-      }
-      ListenableFuture<Query> qf =
-          queryManager.submitQuery(
-              "create function",
-              "create function",
-              "create function",
-              new SubQueryPlan(new SinkRoot(new EOSSource())),
-              workerPlans);
+    /* Validate the command */
+    Pattern pattern = Pattern.compile("(CREATE FUNCTION)([\\s\\S]*)(LANGUAGE SQL;)");
```
senderista: Here and below we are getting into Postgres-specific syntax, so it might be good to call that out at least in comments. The "portable" approach, of course,  would be to add a `createFunction()` method to the `AccessMethod` interface; did we discuss that option?

senderista:
Looks good, I need to refresh my memory though on why we didn't define `AccessMethod.createFunction(`) and have the operator delegate to that. The current design's hard-coding of Postgres UDF syntax in otherwise generic code makes me a bit nervous.

src/edu/washington/escience/myria/coordinator/MasterCatalog.java
senderista: `INSERT OR REPLACE`: https://www.sqlite.org/lang_insert.html

jortiz16:
For now, I only wrote a comment for the function method as you described above. 

If we added a createFunction() method to the access method, what operator would call it? Currently, we are passing function operations through a ```generic-run-SQL``` operator. Using this generic operator forces us to do the validation before passing the function to this operator. If we wanted a createFunction() on the access method, we would need to build a specialized operator and not use the generic one, right?

senderista:
I'm still missing how `createUDF` is fundamentally different enough from `createView` and `createIndex` that it requires a totally different design (i.e., doesn't have a special operator that delegates to the right `AccessMethod` implementation), but I guess since we'll still have to figure out a uniform interface to accommodate Python and Java UDFs, this is fine to get something working.

jortiz16:
So okay to merge for now? I can help refactor this once the Python UDFs are introduced.

senderista:
Sure, let's merge now and revisit after Python UDFs are merged.

PR #842, opened 2016-08-09
Binary ingest support for Booleans and Strings

dhutchis:
This enables support for binary ingest of data with schema containing BOOLEAN or STRING types. The String encoding is variable-width; it reads the length of the string, then the string itself, as expected from `DataOutputStream.readUTF()`.


dhutchis:
Travis says that the files are not formatted properly. Not sure what to do to fix that.

dhutchis:
Formatting solved. @jingjingwang or @jortiz16, this is ready to merge. Want to take a peek? I have used this successfully on the tx-e1 cluster for binary loading of strings.

PR #843, opened 2016-08-10
Add doc on ingesting from curl

dhutchis:
We should document how to stream data into Myria from the command line.

These example scripts are working for us now.

### DIFF
--- a/ingest/README.md
+++ b/ingest/README.md
@@ -1,0 +1,39 @@
+# Myria Ingest from `curl`
+
+The following snippet is useful for streaming data into Myria.
+Modify it to fit your needs, especially the "relationKey" and "schema".
+The `curl` command reads data from stdin. You can pipe data into it from a file or program.
+
+```bash
+TDIR="."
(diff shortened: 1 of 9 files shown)


dhutchis:
The goal of this branch is to dump scripts and misc documentation that we have created during for the ISTC oceanography effort on Amazon and on the TX-E1 cluster, and fish out useful parts later. It is a work-in-progress, to finalize with Jingjing, Tobin, and others, once we decide what would be useful to document and include.

senderista:
I think we should consider a separate `myria-contrib` repo where we could put these sorts of tools, as well as useful MyriaL examples that might not be appropriate for the `myria` or `raco` repos.

@jingjingwang and @jortiz16 can we discuss this at next week's Myria meeting?

jortiz16:
Sure! I agree it should be a separate repo. Also, we might want to show some of these examples on the myria website. In the past, we've talked about adding Myria use cases, but never got around to doing it.

Issue #841, opened 2016-08-09
Handle bad sqllite setup in REST message on ingest


dhutchis:
When sqllite is not working properly, [this line in the `getDatasetStatus` method](../blob/master/src/edu/washington/escience/myria/coordinator/MasterCatalog.java#L1325) could return null.  In this case, the REST call for ingesting a dataset returns a NullPointerException and the user has no idea what went wrong.  I'm still not sure why sqllite is having trouble.  I encountered this when running Myria on my laptop and ingesting a dataset through the [`newDatasetMultipart` method](https://github.com/uwescience/myria/blob/master/src/edu/washington/escience/myria/api/DatasetResource.java#L518).

dhutchis:
Anyone know the cause of this? It popped up again for me, this time on a Myria cluster that used to work fine.  More generally, how do we look at the SQLite that Myria stores catalog information inside?

jortiz16:
Hey Dylan, I would also recommend looking at the catalog and running the query that returns a null response. You can view the catalog by looking for the ```master.catalog``` file under the myria directory and open with ```sqlite3 master.catalog```.

I'm not yet sure why an ingest would return a NullPointer, can you give more details about the error you get here?....If these ingests are failing, then the ```datasetStatus``` would not have anything to return.

dhutchis:
Hi @jortiz16, I can't find the `master.catalog` file in the myria deployment location. I tried `find . -name master.catalog`.

According to the code, the null pointer is a result of the SQLite query not returning any results.

BrandonHaynes:
Irrespective of the underlying cause of this error, it seems like we should be throwing a CatalogException if the relation isn't found in SQLite.

jingjingwang:
Returning null at https://github.com/uwescience/myria/blob/master/src/edu/washington/escience/myria/coordinator/MasterCatalog.java#L1325 is the correct behavior, since we return 404 if the result is null: https://github.com/uwescience/myria/blob/master/src/edu/washington/escience/myria/api/DatasetResource.java#L110. So I don't think the above line is the cause of your `NullPointerException`.

dhutchis:
@jingjingwang This line throws the NullPointerException. It is called when a new dataset is ingested.  

https://github.com/uwescience/myria/blob/master/src/edu/washington/escience/myria/api/DatasetResource.java#L616

jortiz16:
[...]

PR #844, opened 2016-08-11
check null in doIngest.

jingjingwang:
If for any reason we cannot get the dataset status due to a failed ingestion, return a proper error message.

### DIFF
--- a/src/edu/washington/escience/myria/api/DatasetResource.java
+++ b/src/edu/washington/escience/myria/api/DatasetResource.java
@@ -1,41 +1,11 @@
 package edu.washington.escience.myria.api;
 
-import java.io.IOException;
-import java.io.InputStream;
-import java.net.URI;
-import java.util.List;
-import java.util.Set;
-
-import javax.annotation.Nonnull;
-import javax.ws.rs.Consumes;
-import javax.ws.rs.DELETE;
-import javax.ws.rs.GET;
-import javax.ws.rs.POST;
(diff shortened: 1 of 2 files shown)


jingjingwang:
Fixes https://github.com/uwescience/myria/issues/841.

dhutchis:
We should also do something for the deleteDataset line 369.  I had a dataset that appeared to have been deleted but the catalog didn't update.

jingjingwang:
Did you mean https://github.com/uwescience/myria/blob/7644e4ebac6d5f34449960a439fff9b51b4a03a1/src/edu/washington/escience/myria/api/DatasetResource.java#L372? As I read through the code I think this chunk of code is unnecessary since the returned `status` is never used. What do you think? I'm thinking about making `Server.deleteDataset` not return anything.

dhutchis:
If the catalog fails to update then the user should be notified, at least to warn them that the catalog is now in an inconsistent state.

jingjingwang:
I think the user is notified if catalog is not updated since exceptions are thrown, it's only this line https://github.com/uwescience/myria/blob/master/src/edu/washington/escience/myria/parallel/Server.java#L1010 is unnecessary.

dhutchis:
Okay, I see.  Looks good.

dhutchis:
@jingjingwang I altered `deleteDataset` as we discussed to return nothing. If it looks good to you, go ahead and merge.

Thanks for your help Jingjing.

PR #848, opened 2016-08-19
parallel ingest bug fix

jortiz16:
@parmitam  was our first tester and ran into a bug. I realized I forgot to test the last changes we made for more than 2 workers. 

With this fix, I tested for 3 and 5 workers by temporarily adding workers to the SystemTestBase class.

Two fixes:

1) I added a global variable ```byteOffsetFromTruncatedRowAtStart``` which keeps track of the number of bytes we read when skipping over the first partial row. This variable becomes important in cases when a worker needs to reinitialize the parser -- we must have the correct offset otherwise our character skipping will be incorrect. Our 2-worker tests did not cover this because worker 1 never has to worry about this offset when reinitializing the parser (it is always going to be 0) and worker 2 is the last worker so it doesn't need to reinitialize the parser for the last row.

2) This other problem is more subtle. There might be a case when a worker has a partition such as "this_is_an_incomplete_row\n". Our code will read up to the new line ('\n') but we don't have a way to check if there are any other lines to read after that. So I added a way to check if there is something to read. If there isn't, the row is marked as ```flagAsIncomplete```. Here, we still need to check for the quote try/catch in cases as well.

### DIFF
--- a/src/edu/washington/escience/myria/operator/CSVFileScanFragment.java
+++ b/src/edu/washington/escience/myria/operator/CSVFileScanFragment.java
@@ -66,6 +66,7 @@ public class CSVFileScanFragment extends LeafOperator {
   private final long partitionEndByteRange;
 
   private long adjustedStartByteRange;
+  private int byteOffsetFromTruncatedRowAtStart = 0;
   private InputStream partitionInputStream;
   private CSVRecord record;
   private boolean onLastRow;
@@ -269,7 +270,9 @@ public class CSVFileScanFragment extends LeafOperator {
             long characterPositionAtBeginningOfRecord =
                 (record == null) ? 0 : record.getCharacterPosition();
             InputStream completePartitionStream =
-                source.getInputStream(adjustedStartByteRange, finalBytePositionFound);
+                source.getInputStream(
+                    adjustedStartByteRange + byteOffsetFromTruncatedRowAtStart,
+                    finalBytePositionFound);
             BufferedReader reader =
                 new BufferedReader(new InputStreamReader(completePartitionStream));
             reader.skip(characterPositionAtBeginningOfRecord);
@@ -390,12 +393,13 @@ public class CSVFileScanFragment extends LeafOperator {
        */
       if (partitionStartByteRange != 0) {
         int firstChar = partitionInputStream.read();
-        int byteOffset = 1;
+        byteOffsetFromTruncatedRowAtStart = 1;
+
         if (firstChar != '\n' && firstChar != '\r') {
           boolean newLineFound = false;
           while (!newLineFound) {
             int currentChar = partitionInputStream.read();
-            byteOffset++;
+            byteOffsetFromTruncatedRowAtStart++;
             if (currentChar == '\n' || currentChar == '\r' || currentChar == -1) {
               newLineFound = true;
               /*
@@ -406,19 +410,26 @@ public class CSVFileScanFragment extends LeafOperator {
                 flagAsIncomplete = true;
               } else if (currentChar == '\r') {
                 currentChar = partitionInputStream.read();
+                byteOffsetFromTruncatedRowAtStart++;
                 if (currentChar != '\n') {
+                  byteOffsetFromTruncatedRowAtStart--;
                   partitionInputStream =
                       source.getInputStream(
-                          adjustedStartByteRange + byteOffset, partitionEndByteRange);
+                          adjustedStartByteRange + byteOffsetFromTruncatedRowAtStart,
+                          partitionEndByteRange);
                 }
               }
             }
           }
         } else if (firstChar == '\r') {
           int currentChar = partitionInputStream.read();
+          byteOffsetFromTruncatedRowAtStart++;
           if (currentChar != '\n') {
+            byteOffsetFromTruncatedRowAtStart--;
             partitionInputStream =
-                source.getInputStream(adjustedStartByteRange + byteOffset, partitionEndByteRange);
+                source.getInputStream(
+                    adjustedStartByteRange + byteOffsetFromTruncatedRowAtStart,
+                    partitionEndByteRange);
           }
         }
       }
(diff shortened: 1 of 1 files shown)


src/edu/washington/escience/myria/operator/CSVFileScanFragment.java
```diff
@@ -430,6 +441,23 @@ protected void init(final ImmutableMap<String, Object> execEnvVars) throws DbExc
...
+           * last row
+           */
+          if (e.getMessage() != null && e.getMessage().contains(truncatedQuoteErrorMessage)) {
+            onLastRow = true;
+          } else {
+            throw e;
+          }
+        }
+
         for (int i = 0; i < numberOfSkippedLines; i++) {
```
senderista: We can still blow up here if a header line after the first line is truncated within a quote, right? Also, what happens if not all headers are contained within the first partition? And shouldn't the header skipping be conditioned on being the first worker (we'd need to handle the case where header lines begin in subsequent partitions, though)?

jortiz16: Yes, this should be conditioned for the first worker....

But headers in subsequent partitions? How likely would that happen given that we try to keep the size of the partitions at a minimum number (believe it was 100MB). I guess a case where there is a very long header?

senderista: Very unlikely, just possible...

If we're actually going to punt on this then we at least need to mention it in a comment.

src/edu/washington/escience/myria/operator/CSVFileScanFragment.java
```diff
@@ -430,6 +441,23 @@ protected void init(final ImmutableMap<String, Object> execEnvVars) throws DbExc
                 new BufferedReader(new InputStreamReader(partitionInputStream)),
                 CSVFormat.newFormat(delimiter).withQuote(quote).withEscape(escape));
         iterator = parser.iterator();
+
+        try {
+          if (!iterator.hasNext()) {
```
senderista: Is this intended to cover the scenario where the first record is truncated inside a quote? How is that not covered by the `catch` handler for the first call to `iterator.next()` in `fetchNextReady()`?

jortiz16: No, this ```hasNext()``` was intended to cover the case when the worker only has a partial row with a new line (i.e. something like... "example_incomplete_row\n"). 

But I had to add the try/catch because if the partial record had a truncated quote, this ```hasNext()``` would blow up.

senderista: Sorry, I'm still confused. You mean the partial record has an unclosed quoted field followed by a newline at the very end of the partition?

jortiz16: No, sorry, I just picked a really bad example earlier. This ```hasNext()``` case has nothing to do with quotes. I meant the case where a worker _only_ has a partial record ending with a new line. Our logic would read this partial record up to the new line, but then have nothing else to read after. That is the case I'm trying to cover.

In other words, ```hasNext()``` simply makes sure that there is something to read after we've parsed up to the new line.

I was required to add the try/catch because this ```hasNext()``` check could blow up if we have a row with a truncated quote.

senderista: OK, but if you've read all the way to the end of the partition, couldn't you deduce that if `adjustedStartByteRange + byteOffsetFromTruncatedRowAtStart == partitionEndByteRange`, then there is nothing left to read? Alternatively, you could try to read one byte past the newline and set `flagAsIncomplete` if `read()` returned -1. I don't see why it's necessary to call `iterator.hasNext()` to determine this information.

src/edu/washington/escience/myria/operator/CSVFileScanFragment.java
jortiz16: So I think I'll only mention it on a comment for now. I don't immediately see how we can handle the case where other workers (besides the first one) have skipped lines. A worker would only know which lines to skip given the number of lines the previous workers have skipped. This sounds like it would have to be some sort of post-ingest-processing step that would might require worker communication. I don't think we'll be able to do this in parallel if we wanted to handle it.

src/edu/washington/escience/myria/operator/CSVFileScanFragment.java
jortiz16: So I changed this to ```adjustedStartByteRange + byteOffsetFromTruncatedRowAtStart -1 == partitionEndByteRange```

-1 because the ```byteOffsetFromTruncatedRowAtStart``` is counting the bytes inclusively based on the byte range. For example if the ```adjustedStartByteRange``` is 3 and the ```partitionEndByteRange``` is 7, then the ```byteOffsetFromTruncatedRowAtStart``` would be 5.

src/edu/washington/escience/myria/operator/CSVFileScanFragment.java
```diff
@@ -434,6 +433,10 @@ protected void init(final ImmutableMap<String, Object> execEnvVars) throws DbExc
         }
       }
 
+      if (adjustedStartByteRange + byteOffsetFromTruncatedRowAtStart - 1 == partitionEndByteRange) {
```
senderista: Maybe a comment that we've hit the end of the partition?

src/edu/washington/escience/myria/operator/CSVFileScanFragment.java
```diff
@@ -442,25 +445,12 @@ protected void init(final ImmutableMap<String, Object> execEnvVars) throws DbExc
...
-        } catch (Exception e) {
-          /*
-           * FIX ME: If we hit an exception for a malformed row (in case of quotes for example), we mark this as the
-           * last row
-           */
-          if (e.getMessage() != null && e.getMessage().contains(truncatedQuoteErrorMessage)) {
-            onLastRow = true;
-          } else {
-            throw e;
+        /* FIX ME: For now, we only support cases where the numberOfSkippedLines applies to the first worker */
```
senderista: How about "all skipped lines are contained within the first partition"?

PR #849, opened 2016-08-20
encoding for DataOutput operator

jortiz16:
Currently, there's no way to encode the DataOutput operator. We were also missing JSON subtype annotation for UriSink and for the different types of supported TupleWriters.

### DIFF
--- a/src/edu/washington/escience/myria/TupleWriter.java
+++ b/src/edu/washington/escience/myria/TupleWriter.java
@@ -33,7 +33,7 @@ import edu.washington.escience.myria.storage.ReadableTable;
 )
 @JsonSubTypes({
   @Type(name = "CSV", value = CsvTupleWriter.class),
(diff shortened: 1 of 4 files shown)


src/edu/washington/escience/myria/TupleWriter.java
```diff
@@ -22,6 +26,16 @@
...
  *
  */
+@JsonTypeInfo(
+  use = JsonTypeInfo.Id.NAME,
+  include = JsonTypeInfo.As.PROPERTY,
+  property = "writerType"
+)
+@JsonSubTypes({
+  @Type(name = "CSV", value = CsvTupleWriter.class),
+  @Type(name = "Json", value = JsonTupleWriter.class),
```
senderista: IMO should be capitalized since it's an acronym just like CSV.

PR #850, opened 2016-08-20
driverDir CLI; revamp CLI usage printing

dhutchis:
### Background

In order for the DriverLauncher to work, it must have a directory that is writeable by the DriverLauncher and readable by the launched Driver, on whatever node that Yarn allocates the Driver to.  The DriverLauncher places the myria jar and other configuration information inside this location, for transfer to the Driver itself.

The default file system this directory exists in is given by the fs.defaultFS parameter in core-site.xml.  If the defaultFS parameter is not specified, it defaults to file:///.  If you need to change the filesystem (say, to HDFS or to S3). then modify core-site.xml.

The default file location within the filesystem is /tmp.  If you need to change this, then specify the -driverDir parameter on the command line when you start Myria.

*** This will allow you to use Myria without HDFS. ***  (if you have a directory visible to both the driver launcher and the driver, wherever it is launched. For example, this is easy with a Lustre parallel file system.)

### CLI usage

The driver launcher will now log what parameters were expected along with their doc strings, and what parameters were given.

### DIFF
--- a/src/edu/washington/escience/myria/daemon/MyriaDriverLauncher.java
+++ b/src/edu/washington/escience/myria/daemon/MyriaDriverLauncher.java
@@ -250,11 +250,11 @@ public final class MyriaDriverLauncher {
 
                   final StringBuilder sb = new StringBuilder(simpleName);
                   if (!shortName.isEmpty()) {
-                    sb.append(" [-").append(shortName).append("]");
+                    sb.append(String.format(" [-%s]", shortName));
                   }
                   // If desired, the type of the parameter can be obtained from
                   // commandLineConf.getNamedParameters() --> .getSimpleArgName()
-                  sb.append(": ").append(doc).append("\n -> ");
+                  sb.append(String.format(": %s\n-> ", doc));
 
                   boolean ok;
                   try {
@@ -266,9 +266,11 @@ public final class MyriaDriverLauncher {
                     ok = false;
                   }
                   if (!ok) {
-                    if (!defaultVal.isEmpty())
+                    if (!defaultVal.isEmpty()) {
                       sb.append("[cannot parse; using default] ").append(defaultVal);
-                    else sb.append("[cannot parse; no default]");
+                    } else {
+                      sb.append("[cannot parse; no default]");
+                    }
                   }
                   return sb.toString();
                 })
--- a/src/edu/washington/escience/myria/daemon/MyriaDriverLauncher.java
+++ b/src/edu/washington/escience/myria/daemon/MyriaDriverLauncher.java
@@ -222,30 +222,25 @@ public final class MyriaDriverLauncher {
             .map(
                 c -> {
                   final NamedParameter annotation = c.getAnnotation(NamedParameter.class);
-
                   final String fullName = c.getName();
                   final String simpleName = c.getSimpleName();
                   final String shortName = annotation.short_name();
                   final String doc = annotation.doc();
                   final String defaultVal;
-                  {
-                    if (!annotation
-                        .default_value()
-                        .equals(NamedParameter.REEF_UNINITIALIZED_VALUE)) {
-                      defaultVal = annotation.default_value();
-                    } else if (!annotation.default_class().equals(Void.class)) {
-                      defaultVal = annotation.default_class().getSimpleName();
-                    } else if (annotation.default_values().length > 0) {
-                      defaultVal = Arrays.toString(annotation.default_values());
-                    } else if (annotation.default_classes().length > 0) {
-                      final String classNames =
-                          Arrays.stream(annotation.default_classes())
-                              .map(Class::getSimpleName)
-                              .reduce("", (a, b) -> a + ", " + b);
-                      defaultVal = String.format("[%s]", classNames);
-                    } else {
-                      defaultVal = "";
-                    }
+                  if (!annotation.default_value().equals(NamedParameter.REEF_UNINITIALIZED_VALUE)) {
+                    defaultVal = annotation.default_value();
+                  } else if (!annotation.default_class().equals(Void.class)) {
+                    defaultVal = annotation.default_class().getSimpleName();
+                  } else if (annotation.default_values().length > 0) {
+                    defaultVal = Arrays.toString(annotation.default_values());
+                  } else if (annotation.default_classes().length > 0) {
+                    final String classNames =
+                        Arrays.stream(annotation.default_classes())
+                            .map(Class::getSimpleName)
+                            .reduce("", (a, b) -> a + ", " + b);
+                    defaultVal = String.format("[%s]", classNames);
+                  } else {
+                    defaultVal = "";
                   }
 
                   final StringBuilder sb = new StringBuilder(simpleName);
@@ -314,7 +309,6 @@ public final class MyriaDriverLauncher {
       final String configPath = commandLineInjector.getNamedInstance(ConfigPath.class);
       final String javaLibPath = commandLineInjector.getNamedInstance(JavaLibPath.class);
       final String nativeLibPath = commandLineInjector.getNamedInstance(NativeLibPath.class);
-
       final Configuration globalConf = getMyriaGlobalConf(configPath);
       final String serializedGlobalConf = new AvroConfigurationSerializer().toString(globalConf);
       final Configuration globalConfWrapper =
--- a/src/edu/washington/escience/myria/daemon/MyriaDriverLauncher.java
+++ b/src/edu/washington/escience/myria/daemon/MyriaDriverLauncher.java
@@ -9,6 +9,7 @@ import java.nio.file.Files;
 import java.nio.file.Path;
 import java.nio.file.Paths;
 import java.util.ArrayList;
+import java.util.Arrays;
 import java.util.List;
 
 import javax.annotation.Nullable;
@@ -55,9 +56,6 @@ public final class MyriaDriverLauncher {
 
   private static final Logger LOGGER = LoggerFactory.getLogger(MyriaDriverLauncher.class);
 
-  private static final String USAGE_STRING =
-      "Usage: MyriaDriverLauncher [-runtimeClass <runtimeClass>] [-driverDir <driverDir>] -configPath <configPath> -javaLibPath <javaLibPath> -nativeLibPath <nativeLibPath>";
-
   private static final long DRIVER_PING_TIMEOUT_MILLIS = 60 * 1000;
 
   /**
@@ -65,7 +63,8 @@ public final class MyriaDriverLauncher {
    * @throws Exception if the Driver can't start.
    */
   public static void main(final String[] args) throws Exception {
-    run(args);
+    LauncherStatus run = run(args);
+    LOGGER.info("Exit with status " + run);
   }
 
   private final REEF reef;
@@ -138,7 +137,9 @@ public final class MyriaDriverLauncher {
             .set(DriverConfiguration.ON_CLIENT_CLOSED, MyriaDriver.ClientCloseHandler.class);
 
     if (driverJobSubmissionDirectory != null) {
-      driverConf = driverConf.set(DriverConfiguration.DRIVER_JOB_SUBMISSION_DIRECTORY, driverJobSubmissionDirectory);
+      driverConf =
+          driverConf.set(
+              DriverConfiguration.DRIVER_JOB_SUBMISSION_DIRECTORY, driverJobSubmissionDirectory);
     }
     for (String dirPath : libPaths) {
       for (String filePath : getFileNamesInDirectory(Paths.get(dirPath))) {
@@ -213,9 +214,60 @@ public final class MyriaDriverLauncher {
   }
 
   /**
+   * Prints the expected configuration, compared to the actual configuration given.
+   * Used for logging command line arguments.
+   */
+  private static void logStartupMessage(Class<? extends Name<?>>[] classes, Injector inj) {
+    final StringBuilder sb = new StringBuilder("MyriaDriverLauncher configuration:\n");
+    for (int i = 0; i < classes.length; i++) {
+      final Class<? extends Name<?>> c = classes[i];
+      final NamedParameter annotation = c.getAnnotation(NamedParameter.class);
+
+      final String fullName = c.getName();
+      final String simpleName = c.getSimpleName();
+      final String shortName = annotation.short_name();
+      final String doc = annotation.doc();
+      final String defaultVal;
+      {
+        if (!annotation.default_value().equals(NamedParameter.REEF_UNINITIALIZED_VALUE))
+          defaultVal = annotation.default_value();
+        else if (!annotation.default_class().equals(Void.class))
+          defaultVal = annotation.default_class().getSimpleName();
+        else if (annotation.default_values().length > 0)
+          defaultVal = Arrays.toString(annotation.default_values());
+        else if (annotation.default_classes().length > 0) {
+          Class<?>[] default_classes = annotation.default_classes();
+          String[] sss = new String[default_classes.length];
+          for (int i1 = 0; i1 < default_classes.length; i1++) {
+            sss[i1] = default_classes[i1].getSimpleName();
+          }
+          defaultVal = Arrays.toString(sss);
+        } else defaultVal = "";
+      }
+
+      if (i > 0) sb.append('\n');
+      sb.append(simpleName);
+      if (!shortName.isEmpty()) sb.append(" [-").append(shortName).append("]");
+
+      //      sb.append(" (").append(expectedParameter.getSimpleArgName()).append("): ")
+      sb.append(": ").append(doc).append("\n -> ");
+
+      boolean ok;
+      try {
+        if ((ok = inj.isParameterSet(fullName))) sb.append(inj.getInstance(fullName).toString());
+      } catch (InjectionException e) {
+        ok = false;
+      }
+      if (!ok)
+        if (!defaultVal.isEmpty()) sb.append("[cannot parse; using default] ").append(defaultVal);
+        else sb.append("[cannot parse; no default]");
+    }
+    LOGGER.info(sb.toString());
+  }
+
+  /**
    * Launch the Myria driver.
    *
-   * @param runtimeConf The runtime configuration (e.g. Local, YARN, etc)
    * @param args Command line arguments.
    * @throws InjectionException
    * @throws java.io.IOException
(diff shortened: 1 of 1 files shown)


dhutchis:
@senderista This should be good to go.

src/edu/washington/escience/myria/daemon/MyriaDriverLauncher.java
```diff
@@ -312,6 +384,16 @@ public synchronized void close() {
   public static final class RuntimeClassName implements Name<String> {}
 
   /**
+   * Command line parameter: runtime configuration class to use (defaults to local runtime).
```
senderista: fix comment

src/edu/washington/escience/myria/daemon/MyriaDriverLauncher.java
```diff
@@ -212,9 +214,60 @@ private static int getDriverMemory(final Configuration conf) throws InjectionExc
...
+          }
+          defaultVal = Arrays.toString(sss);
+        } else defaultVal = "";
+      }
+
+      if (i > 0) sb.append('\n');
+      sb.append(simpleName);
+      if (!shortName.isEmpty()) sb.append(" [-").append(shortName).append("]");
+
+      //      sb.append(" (").append(expectedParameter.getSimpleArgName()).append("): ")
```
senderista: let's not leave commented-out code in

src/edu/washington/escience/myria/daemon/MyriaDriverLauncher.java
```diff
@@ -212,9 +214,60 @@ private static int getDriverMemory(final Configuration conf) throws InjectionExc
   }
 
   /**
+   * Prints the expected configuration, compared to the actual configuration given.
+   * Used for logging command line arguments.
+   */
+  private static void logStartupMessage(Class<? extends Name<?>>[] classes, Injector inj) {
```
senderista: I'd prefer a method that didn't mix computation and side effects. How about this method just generates the log message and the caller logs it?

src/edu/washington/escience/myria/daemon/MyriaDriverLauncher.java
```diff
@@ -54,21 +56,15 @@
...
    * @param args full path to the configuration directory.
    * @throws Exception if the Driver can't start.
    */
   public static void main(final String[] args) throws Exception {
-    if (!(args.length == 6 || args.length == 8)) {
-      System.err.println(USAGE_STRING);
-      System.exit(-1);
-    }
-    run(args);
+    LauncherStatus run = run(args);
```
senderista: `run` -> `status`

src/edu/washington/escience/myria/daemon/MyriaDriverLauncher.java
```diff
@@ -212,9 +214,60 @@ private static int getDriverMemory(final Configuration conf) throws InjectionExc
...
+      if (i > 0) sb.append('\n');
+      sb.append(simpleName);
+      if (!shortName.isEmpty()) sb.append(" [-").append(shortName).append("]");
+
+      //      sb.append(" (").append(expectedParameter.getSimpleArgName()).append("): ")
+      sb.append(": ").append(doc).append("\n -> ");
+
+      boolean ok;
+      try {
+        if ((ok = inj.isParameterSet(fullName))) sb.append(inj.getInstance(fullName).toString());
```
senderista: We don't use assignments in conditionals, and we always use braces in conditionals.

src/edu/washington/escience/myria/daemon/MyriaDriverLauncher.java
```diff
@@ -212,9 +214,60 @@ private static int getDriverMemory(final Configuration conf) throws InjectionExc
   }
 
   /**
+   * Prints the expected configuration, compared to the actual configuration given.
+   * Used for logging command line arguments.
+   */
+  private static void logStartupMessage(Class<? extends Name<?>>[] classes, Injector inj) {
+    final StringBuilder sb = new StringBuilder("MyriaDriverLauncher configuration:\n");
+    for (int i = 0; i < classes.length; i++) {
```
senderista: If you're not using the loop index elsewhere, you should be able to replace this with a foreach loop.

src/edu/washington/escience/myria/daemon/MyriaDriverLauncher.java
```diff
@@ -212,9 +214,60 @@ private static int getDriverMemory(final Configuration conf) throws InjectionExc
...
+      final Class<? extends Name<?>> c = classes[i];
+      final NamedParameter annotation = c.getAnnotation(NamedParameter.class);
+
+      final String fullName = c.getName();
+      final String simpleName = c.getSimpleName();
+      final String shortName = annotation.short_name();
+      final String doc = annotation.doc();
+      final String defaultVal;
+      {
+        if (!annotation.default_value().equals(NamedParameter.REEF_UNINITIALIZED_VALUE))
```
senderista: braces around all conditional blocks

src/edu/washington/escience/myria/daemon/MyriaDriverLauncher.java
```diff
@@ -212,9 +214,60 @@ private static int getDriverMemory(final Configuration conf) throws InjectionExc
...
+          defaultVal = annotation.default_value();
+        else if (!annotation.default_class().equals(Void.class))
+          defaultVal = annotation.default_class().getSimpleName();
+        else if (annotation.default_values().length > 0)
+          defaultVal = Arrays.toString(annotation.default_values());
+        else if (annotation.default_classes().length > 0) {
+          Class<?>[] default_classes = annotation.default_classes();
```
senderista: This is a pretty clunky, C-style way to create an array of default values. Couldn't you just iterate with a foreach loop over `annotation.default_classes()` and append the results of `getSimpleName()` one-at-a-time to a `StringBuilder`?

src/edu/washington/escience/myria/daemon/MyriaDriverLauncher.java
```diff
@@ -212,9 +214,60 @@ private static int getDriverMemory(final Configuration conf) throws InjectionExc
...
+          String[] sss = new String[default_classes.length];
+          for (int i1 = 0; i1 < default_classes.length; i1++) {
+            sss[i1] = default_classes[i1].getSimpleName();
+          }
+          defaultVal = Arrays.toString(sss);
+        } else defaultVal = "";
+      }
+
+      if (i > 0) sb.append('\n');
```
senderista: You could avoid using an array index here by using Guava's `Joiner` instead of `StringBuilder`.

src/edu/washington/escience/myria/daemon/MyriaDriverLauncher.java
```diff
@@ -212,9 +214,60 @@ private static int getDriverMemory(final Configuration conf) throws InjectionExc
...
+      final String shortName = annotation.short_name();
+      final String doc = annotation.doc();
+      final String defaultVal;
+      {
+        if (!annotation.default_value().equals(NamedParameter.REEF_UNINITIALIZED_VALUE))
+          defaultVal = annotation.default_value();
+        else if (!annotation.default_class().equals(Void.class))
+          defaultVal = annotation.default_class().getSimpleName();
+        else if (annotation.default_values().length > 0)
+          defaultVal = Arrays.toString(annotation.default_values());
```
senderista: What does the output of `Arrays.toString()` look like? If we want to control the output, we should specify the delimiter using Guava's `Joiner` or the like.

src/edu/washington/escience/myria/daemon/MyriaDriverLauncher.java
```diff
@@ -212,9 +214,60 @@ private static int getDriverMemory(final Configuration conf) throws InjectionExc
...
+          for (int i1 = 0; i1 < default_classes.length; i1++) {
+            sss[i1] = default_classes[i1].getSimpleName();
+          }
+          defaultVal = Arrays.toString(sss);
+        } else defaultVal = "";
+      }
+
+      if (i > 0) sb.append('\n');
+      sb.append(simpleName);
+      if (!shortName.isEmpty()) sb.append(" [-").append(shortName).append("]");
```
senderista: Better to do this at a higher level with `String.format()` and `Joiner`.

src/edu/washington/escience/myria/daemon/MyriaDriverLauncher.java
```diff
@@ -226,40 +279,59 @@ private static int getDriverMemory(final Configuration conf) throws InjectionExc
...
+            ConfigPath.class,
+            JavaLibPath.class,
+            NativeLibPath.class
+          };
+      final Configuration commandLineConf =
+          CommandLine.parseToConfiguration(args, commandLineClasses);
+      final Injector commandLineInjector = tang.newInjector(commandLineConf);
+      logStartupMessage(commandLineClasses, commandLineInjector);
+      final String runtimeClassName = commandLineInjector.getNamedInstance(RuntimeClassName.class);
+      final String driverJobSubmissionDirectory =
```
senderista: Let's get rid of this ternary operator for clarity:
```
String driverJobSubmissionDirectory = null;
if (commandLineInjector.isParameterSet(DriverJobSubmissionDirectory.class)) {
  driverJobSubmissionDirectory = commandLineInjector.getNamedInstance(DriverJobSubmissionDirectory.class);
}
```

src/edu/washington/escience/myria/daemon/MyriaDriverLauncher.java
```diff
@@ -212,9 +214,60 @@ private static int getDriverMemory(final Configuration conf) throws InjectionExc
...
+      sb.append(simpleName);
+      if (!shortName.isEmpty()) sb.append(" [-").append(shortName).append("]");
+
+      //      sb.append(" (").append(expectedParameter.getSimpleArgName()).append("): ")
+      sb.append(": ").append(doc).append("\n -> ");
+
+      boolean ok;
+      try {
+        if ((ok = inj.isParameterSet(fullName))) sb.append(inj.getInstance(fullName).toString());
+      } catch (InjectionException e) {
```
senderista: I don't see why we need the flag at all. Why not just do the logging here in the `catch` block?

senderista: Why would we expect to catch `InjectionException` on calling `Injector.getInstance()` if `Injector.isParameterSet()` succeeded?

src/edu/washington/escience/myria/daemon/MyriaDriverLauncher.java
```diff
@@ -212,9 +214,60 @@ private static int getDriverMemory(final Configuration conf) throws InjectionExc
...
+        ok = false;
```
senderista: If we don't need the `catch` block above, then we don't need the flag.

src/edu/washington/escience/myria/daemon/MyriaDriverLauncher.java
```diff
@@ -226,40 +279,59 @@ private static int getDriverMemory(final Configuration conf) throws InjectionExc
...
+                  driverMemoryMB,
+                  new String[] {javaLibPath},
+                  new String[] {nativeLibPath}),
+              globalConfWrapper);
+
+      return tang.newInjector(getRuntimeConf(runtimeClassName), getClientConf())
+          .getInstance(MyriaDriverLauncher.class)
+          .run(driverConf);
+    } catch (ParseException | InjectionException e) {
+      LOGGER.error("Problem with command line options", e);
```
senderista: If any command-line misconfiguration should have been diagnosed in the previous log message, then should we direct the user to that message ("see above" etc.)?

src/edu/washington/escience/myria/daemon/MyriaDriverLauncher.java
dhutchis: I am using the loop index below.

src/edu/washington/escience/myria/daemon/MyriaDriverLauncher.java
dhutchis: It looks like `[str1, str2, str3]` for an array of three strings

src/edu/washington/escience/myria/daemon/MyriaDriverLauncher.java
dhutchis: Is it a style thing to have final variables everywhere?

dhutchis:
Addressed comments

src/edu/washington/escience/myria/daemon/MyriaDriverLauncher.java
senderista: Oh, if you were wondering why `driverJobSubmissionDirectory` wasn't `final` in my example, it's just because I was reassigning it in the conditional. I think that's acceptable given the added readability of the conditional.

src/edu/washington/escience/myria/daemon/MyriaDriverLauncher.java
```diff
@@ -297,12 +301,17 @@ public static LauncherStatus run(final String[] args)
...
           CommandLine.parseToConfiguration(args, commandLineClasses);
       final Injector commandLineInjector = tang.newInjector(commandLineConf);
-      logStartupMessage(commandLineClasses, commandLineInjector);
+      LOGGER.info(genStartupMessage(commandLineClasses, commandLineInjector));
       final String runtimeClassName = commandLineInjector.getNamedInstance(RuntimeClassName.class);
-      final String driverJobSubmissionDirectory =
-          commandLineInjector.isParameterSet(DriverJobSubmissionDirectory.class)
-              ? commandLineInjector.getNamedInstance(DriverJobSubmissionDirectory.class)
-              : null;
+      final String driverJobSubmissionDirectory;
```
senderista: Not sure why you introduced an extra block here. I would just say (foregoing the `final` for readability):
```
String driverJobSubmissionDirectory = null;
if (commandLineInjector.isParameterSet(DriverJobSubmissionDirectory.class)) {
  driverJobSubmissionDirectory = commandLineInjector.getNamedInstance(DriverJobSubmissionDirectory.class);
}
```

src/edu/washington/escience/myria/daemon/MyriaDriverLauncher.java
```diff
@@ -229,55 +228,60 @@ private static void logStartupMessage(Class<? extends Name<?>>[] classes, Inject
...
           defaultVal = annotation.default_class().getSimpleName();
-        else if (annotation.default_values().length > 0)
+        } else if (annotation.default_values().length > 0) {
           defaultVal = Arrays.toString(annotation.default_values());
-        else if (annotation.default_classes().length > 0) {
-          Class<?>[] default_classes = annotation.default_classes();
-          String[] sss = new String[default_classes.length];
-          for (int i1 = 0; i1 < default_classes.length; i1++) {
-            sss[i1] = default_classes[i1].getSimpleName();
+        } else if (annotation.default_classes().length > 0) {
```
senderista: I think this code could be shortened/clarified quite a bit with some Guava/Java 8 constructs (untested):
```
final String[] classNames = Arrays.stream(annotation.default_classes()).map(c -> c.getSimpleName()).toArray();
final String classNameMsg = String.format("[%s]", Joiner.on(", ").join(classNames));
```

src/edu/washington/escience/myria/daemon/MyriaDriverLauncher.java
```diff
@@ -229,55 +228,60 @@ private static void logStartupMessage(Class<? extends Name<?>>[] classes, Inject
...
       sb.append(": ").append(doc).append("\n -> ");
 
       boolean ok;
       try {
-        if ((ok = inj.isParameterSet(fullName))) sb.append(inj.getInstance(fullName).toString());
+        ok = inj.isParameterSet(fullName);
+        if (ok) {
+          sb.append(inj.getInstance(fullName).toString());
+        }
       } catch (InjectionException e) {
```
senderista: Again, is it possible for `InjectionException` to be thrown by `getInstance()` if `isParameterSet()` has succeeded? If not then we don't need the flag (or the `catch` block: just an `if`/`else` statement suffices). (From inspecting source it appears that `isParameterSet()` can only throw `BindException`.)

src/edu/washington/escience/myria/daemon/MyriaDriverLauncher.java
```diff
@@ -229,55 +228,60 @@ private static void logStartupMessage(Class<? extends Name<?>>[] classes, Inject
...
+            }
+            b.append(", ");
           }
-          defaultVal = Arrays.toString(sss);
+          defaultVal = b.toString();
         } else defaultVal = "";
       }
 
-      if (i > 0) sb.append('\n');
+      if (i > 0) {
```
senderista: This strongly suggests that you can just append each line to a list and generate a newline-separated message with `Joiner.on("\n")`. That would help expose intention (high-level structure of the message) rather than mechanism (appending each delimiter one-at-a-time to a `StringBuilder`).

src/edu/washington/escience/myria/daemon/MyriaDriverLauncher.java
dhutchis: Cool.  I haven't used the Java 8 lambda stream stuff before.

src/edu/washington/escience/myria/daemon/MyriaDriverLauncher.java
dhutchis: My guess is that the parameter could be set, but a bad parameter could be provided.  For example, a string "abc" could be provided where an int parameter is expected. Then the InjectionException might get thrown.  I can't tell by inspecting the source code; it's complicated.  Instead I am being proactively defensive.

Since you mentioned the BindException, I will add that in.

src/edu/washington/escience/myria/daemon/MyriaDriverLauncher.java
dhutchis: Hmm.  If I see code full of final variables and then encounter a non-final one, I would be suspicious. I don't want to give future developers that interpretation.

senderista: Sure, it's a matter of taste whether you prefer consistency or (IMO) readability. I certainly try to use `final` wherever I can.

src/edu/washington/escience/myria/daemon/MyriaDriverLauncher.java
senderista: `<soapbox>`Unexpected exceptions shouldn't be caught (except in special cases). "Defensive coding" often just masks bugs. I'd much rather debug an unexpected crash with a nice stack trace than an obscure data corruption bug.`</soapbox>`

Also, I don't think Tang would have let a parameter type mismatch escape the `Injector` instantiation. That said, you are just printing a log message here, so excessive paranoia isn't called for.

src/edu/washington/escience/myria/daemon/MyriaDriverLauncher.java
senderista: I haven't either, TBH :) We just moved to Java 8 a few months ago and I'm trying to maintain consistency with existing style, but the REEF-specific code already uses lambdas all over the place anyway.

src/edu/washington/escience/myria/daemon/MyriaDriverLauncher.java
```diff
@@ -217,65 +217,64 @@ private static int getDriverMemory(final Configuration conf) throws InjectionExc
...
+                    ok = inj.isParameterSet(fullName);
+                    if (ok) {
+                      sb.append(inj.getInstance(fullName).toString());
+                    }
+                  } catch (InjectionException | BindException e) {
+                    ok = false;
+                  }
+                  if (!ok) {
+                    if (!defaultVal.isEmpty())
+                      sb.append("[cannot parse; using default] ").append(defaultVal);
```
senderista: braces

src/edu/washington/escience/myria/daemon/MyriaDriverLauncher.java
```diff
@@ -217,65 +217,64 @@ private static int getDriverMemory(final Configuration conf) throws InjectionExc
...
+                    else sb.append("[cannot parse; no default]");
+                  }
+                  return sb.toString();
+                })
+            .reduce("", (a, b) -> a + "\n" + b);
```
senderista: I personally think that using `Joiner` with a `List` is clearer than `reduce`, but this is undeniably cool :)

src/edu/washington/escience/myria/daemon/MyriaDriverLauncher.java
```diff
@@ -217,65 +217,64 @@ private static int getDriverMemory(final Configuration conf) throws InjectionExc
...
+                              .reduce("", (a, b) -> a + ", " + b);
+                      defaultVal = String.format("[%s]", classNames);
+                    } else {
+                      defaultVal = "";
+                    }
+                  }
+
+                  final StringBuilder sb = new StringBuilder(simpleName);
+                  if (!shortName.isEmpty()) {
+                    sb.append(" [-").append(shortName).append("]");
```
senderista: `String.format(" [-%s]", shortName)`

src/edu/washington/escience/myria/daemon/MyriaDriverLauncher.java
```diff
@@ -217,65 +217,64 @@ private static int getDriverMemory(final Configuration conf) throws InjectionExc
...
+                  }
+                  // If desired, the type of the parameter can be obtained from
+                  // commandLineConf.getNamedParameters() --> .getSimpleArgName()
+                  sb.append(": ").append(doc).append("\n -> ");
```
senderista: `String.format(": %s\n-> ", doc)`

src/edu/washington/escience/myria/daemon/MyriaDriverLauncher.java
```diff
@@ -212,59 +213,138 @@ private static int getDriverMemory(final Configuration conf) throws InjectionExc
...
+            .map(
+                c -> {
+                  final NamedParameter annotation = c.getAnnotation(NamedParameter.class);
+
+                  final String fullName = c.getName();
+                  final String simpleName = c.getSimpleName();
+                  final String shortName = annotation.short_name();
+                  final String doc = annotation.doc();
+                  final String defaultVal;
+                  {
```
senderista: superfluous block

src/edu/washington/escience/myria/daemon/MyriaDriverLauncher.java
```diff
@@ -212,59 +213,138 @@ private static int getDriverMemory(final Configuration conf) throws InjectionExc
...
+   * Prints the expected configuration, compared to the actual configuration given.
+   * Used for logging command line arguments.
+   */
+  private static String genStartupMessage(Class<? extends Name<?>>[] classes, Injector inj) {
+    String allparams =
+        Arrays.stream(classes)
+            .map(
+                c -> {
+                  final NamedParameter annotation = c.getAnnotation(NamedParameter.class);
+
```
senderista: remove blank line

src/edu/washington/escience/myria/daemon/MyriaDriverLauncher.java
```diff
@@ -212,59 +213,138 @@ private static int getDriverMemory(final Configuration conf) throws InjectionExc
...
+      if (commandLineInjector.isParameterSet(DriverJobSubmissionDirectory.class)) {
+        driverJobSubmissionDirectory =
+            commandLineInjector.getNamedInstance(DriverJobSubmissionDirectory.class);
+      } else {
+        driverJobSubmissionDirectory = null;
+      }
+      final String configPath = commandLineInjector.getNamedInstance(ConfigPath.class);
+      final String javaLibPath = commandLineInjector.getNamedInstance(JavaLibPath.class);
+      final String nativeLibPath = commandLineInjector.getNamedInstance(NativeLibPath.class);
+
```
senderista: remove blank line

Issue #814, opened 2016-02-24
Implement `howPartitioned.isBroadcast` attribute in Myria catalog


senderista:
We need to persist whether a relation has been broadcast to all workers in order for Raco to push joins into `DbQueryScan` when one input is a broadcast relation.

jingjingwang:
My current thought of this problem is as follows. We are relying on two "functions" when distributing relations: a thing called `partitionFunction` that maps one tuple into a partition, and a thing called `cellPartition` that maps a partition to a set of workers (check `GenericShuffleProducer`). `cellPartition` is a one-to-one mapping function in all the cases except `BroadcastProducer` and `HyperShuffleProducer`.

I'd like to merge these two functions into one, which maps a tuple to a set of "destinations" (I don't want to use the word "workers" here since I'd like to stay in the logical world for now). Let's call this new function as "distributeFunction" for now. The # of destinations is a shared property that all subclasses of `distributeFunction` should support. Then `BroadcastProducer` has a `distributeFunction` that maps a tuple to the set of {1, 2, ... # of destinations}. 

After this refactoring, we can simply rely on the JSON encoding/decoding of a distributeFunction to read from/write to catalog, query plan, etc. It also makes the interfaces in `GenericShuffleProducer` simpler, since the only difference is the `distributeFunction`.

senderista:
@jingjingwang makes sense to me. @bmyerz what do you think?

bmyerz:
It makes sense to me functionally, but the design implications are not yet clear to me.

- how do we write the distributeFunction? Is there a grammar for writing it that includes the numdestinations variable and an opaque hash function?

- in Raco, I think it is most straightforward to have fairly high level encodings (e.g., hash_partitioned on what attributes / broadcasted) for use in the rules. I can vaguely envision a more expressive algebra but I don't want to invent it until needed.

Does this make sense?

senderista:
Given recent user interest in this feature, it sounds like we should revisit this and address @bmyerz's concerns about Raco representation.

bmyerz:
If the main scenarios are the two: replicated table and partitioned table, then I would vote in favor of high level encodings in the catalog:

> in Raco, I think it is most straightforward to have fairly high level encodings (e.g., hash_partitioned on what attributes / broadcasted) for use in the rules. I can vaguely envision a more expressive algebra but I don't want to invent it until needed.

jingjingwang:
[...]

Issue #773, opened 2015-09-17
Refactor SingleFieldHashPartitionFunction and MultiFieldHashPartitionFunction into new type AttributeHashPartitionFunction


senderista:
There doesn't seem to be a good reason to have both SingleFieldHashPartitionFunction and MultiFieldHashPartitionFunction, since the former is just a special case of the latter. The distinction seems to be purely for historical reasons (SingleFieldHashPartitionFunction came first).

PR #863, opened 2016-12-19
Add broadcast

jingjingwang:
This turns out to be a big PR including several things:

- Implement DistributeFunction as in #814. A DistributeFunction consists of a PartitionFunction, which maps each tuple into one partition, and mappings from partitions to destinations. Broadcast is one of the distribute functions.
- Record how a dataset is distributed in HowDistributed. This class takes a DistributeFunction and a list of workers. It is both kept in catalog and returned as a part of dataset status.
- Simplify many Producer/Consumer interfaces. Due to the new DistributeFunction interfaces, Several Producer/Consumer subclasses are removed, and several APIs/methods are simplified.
- Also solve #773 together by merging both SingleFieldHash and MultiFieldHash into HashDistributeFunction.

A RACO PR will be following once the reviewing is mostly done.


jingjingwang:
Sorry about mixing all these in one commit @senderista , but I found it to be hard to split since many changes are interleaved.

src/edu/washington/escience/myria/storage/TupleBatch.java
```diff
@@ -15,15 +15,12 @@
...
+import edu.washington.escience.myria.operator.network.distribute.PartitionFunction;
 import edu.washington.escience.myria.proto.TransportProto.TransportMessage;
 import edu.washington.escience.myria.util.IPCUtils;
 import net.jcip.annotations.ThreadSafe;
 
-/**
- * Container class for a batch of tuples. The goal is to amortize memory management overhead.
- *
- */
+/** Container class for a batch of tuples. The goal is to amortize memory management overhead. */
```
senderista: maybe also amortize processing overhead (from code/data locality)?

src/edu/washington/escience/myria/util/HashUtils.java
```diff
@@ -96,9 +94,23 @@ public static int hashValue(
...
+    return hashSubRow(table, hashColumns, row, 0);
+  }
+
+  /**
+   * Compute the hash code of the specified columns in the specified row of the given table.
+   *
+   * @param table the table containing the values to be hashed
+   * @param hashColumns the columns to be hashed. Order matters
+   * @param row the row containing the values to be hashed
+   * @param seedIndex the index of the chosen hashcode
```
senderista: shouldn't this be "index of the chosen hash function"?

test/edu/washington/escience/myria/util/TestUtils.java
```diff
@@ -456,54 +447,50 @@ public static TupleBatchBuffer generateRandomTuples(
...
 
-  /**
-   * Construct a SubQuery that will insert the given tuples (starting on the master) on the specified workers using the
+  /** Construct a SubQuery that will insert the given tuples (starting on the master) on the specified workers using the
    * specified relation key and partition function.
    *
    * @param masterSource the source of tuples, from the master.
    * @param dest the name of the relation into which tuples will be inserted (using overwrite!).
-   * @param pf how tuples will be partitioned on the cluster.
+   * @param df
```
senderista: "how tuples will be distributed on the cluster."

test/edu/washington/escience/myria/util/TestUtils.java
```diff
@@ -456,54 +447,50 @@ public static TupleBatchBuffer generateRandomTuples(
...
```
senderista: "how tuples will be distributed on the cluster."

src/edu/washington/escience/myria/operator/network/distribute/HyperCubeDistributeFunction.java
```diff
@@ -0,0 +1,45 @@
...
+      @JsonProperty("mappedHCDimensions") final int[] mappedHCDimensions,
+      @JsonProperty("callPartition") final List<List<Integer>> cellPartition) {
+    super(
+        new HyperCubePartitionFunction(hyperCubeDimensions, hashedColumns, mappedHCDimensions),
+        cellPartition);
+  }
+
+  /** @return all destinations */
+  public List<Integer> getAllDestinations() {
+    List<Integer> values = new ArrayList<Integer>();
```
jingjingwang: I agree, however I was following the implementation in https://github.com/uwescience/myria/blob/master/src/edu/washington/escience/myria/api/encoding/HyperShuffleProducerEncoding.java#L57. `cellPartition` contains worker IDs and once flattened and sorted (without dedup), it is expected to be [0,1,...n-1].

src/edu/washington/escience/myria/operator/network/distribute/HyperCubePartitionFunction.java
```diff
@@ -0,0 +1,57 @@
...
+      for (int j = 0; j < hashedColumns.length; j++) {
+        p +=
+            Math.floorMod(
+                HashUtils.hashSubRow(tb, new int[] {hashedColumns[j]}, i, mappedHCDimensions[j]),
+                hyperCubeDimensions[mappedHCDimensions[j]]);
+        if (p != hashedColumns.length - 1) {
+          p *= hyperCubeDimensions[mappedHCDimensions[j]];
+        }
+      }
+      partitions[p].set(i);
```
jingjingwang: Similar to above, this follows https://github.com/uwescience/myria/blob/master/src/edu/washington/escience/myria/operator/network/partition/MFMDHashPartitionFunction.java#L50. I'm not sure if I misunderstood anything though.

src/edu/washington/escience/myria/operator/network/distribute/IdentityPartitionFunction.java
```diff
@@ -0,0 +1,42 @@
...
+   */
+  public IdentityPartitionFunction(final int index) {
+    this.index = index;
+  }
+
+  @Override
+  public TupleBatch[] partition(@Nonnull final TupleBatch tb) {
+    BitSet[] partitions = new BitSet[numPartition()];
+    for (int i = 0; i < tb.numTuples(); i++) {
+      partitions[tb.getInt(index, i) - 1].set(i);
```
jingjingwang: We don't have the schema until this method is called, so I don't think we can do it early unless we do it in the Producer: when its schema is set, we check if its distribute function is an 'IdentityDistributeFunction`. An out-of-bound value in the "partition" column will trigger an IndexOutOfBoundException for the array `partitions`, but not making BitSets grow since their sizes are set by `numTuples()` not `numPartitions()`.

src/edu/washington/escience/myria/api/encoding/HyperCubeShuffleProducerEncoding.java
```diff
@@ -0,0 +1,44 @@
...
+  /** distribute function. */
+  @Required public HyperCubeDistributeFunction distributeFunction;
+
+  @Override
+  public GenericShuffleProducer construct(ConstructArgs args) throws MyriaApiException {
+    return new GenericShuffleProducer(
+        null,
+        getRealOperatorIds().toArray(new ExchangePairID[] {}),
+        MyriaUtils.integerSetToIntArray(
+            args.getServer().getRandomWorkers(distributeFunction.getAllDestinations().size())),
```
jingjingwang: Similarly, this follows https://github.com/uwescience/myria/blob/master/src/edu/washington/escience/myria/api/encoding/HyperShuffleProducerEncoding.java#L52. `getAllDestinations()` is simply a helper function, we need the list of worker IDs for the checking in `validateExtra()`.

src/edu/washington/escience/myria/api/encoding/HyperCubeShuffleProducerEncoding.java
```diff
@@ -0,0 +1,44 @@
...
+        distributeFunction);
+  }
+
+  @Override
+  protected void validateExtra() {
```
jingjingwang: See above.

src/edu/washington/escience/myria/coordinator/MasterCatalog.java
```diff
@@ -129,7 +129,7 @@
           + "    program_name TEXT NOT NULL,\n"
           + "    relation_name TEXT NOT NULL,\n"
           + "    num_shards INTEGER NOT NULL,\n"
-          + "    how_partitioned TEXT NOT NULL,\n"
+          + "    how_distributed TEXT NOT NULL,\n"
```
jingjingwang: Yes, and we may need to mark some as "unknown" since we have changed the internal representation of this field.

src/edu/washington/escience/myria/operator/network/LocalMultiwayProducer.java
```diff
@@ -1,39 +1,23 @@
...
-        MyriaArrayUtils.create2DHorizontalIndex(operatorIDs.length),
         new int[] {IPCConnectionPool.SELF_IPC_ID},
-        new FixValuePartitionFunction(0),
-        false);
-  }
-
-  @Override
-  protected TupleBatch[] getTupleBatchPartitions(final TupleBatch tup) {
-    return new TupleBatch[] {tup};
+        new BroadcastDistributeFunction(operatorIDs.length));
```
jingjingwang: There are multiple destinations since there are multiple operator IDs.

src/edu/washington/escience/myria/operator/network/RecoverProducer.java
```diff
@@ -29,13 +25,11 @@
   /** the channel index that this operator is recovering for. */
   private final int channelIndx;
 
-  /**
-   * @param child the child who provides data for this producer to distribute.
+  /** @param child the child who provides data for this producer to distribute.
```
jingjingwang: I changed the other two but kept `oriProducer'.

src/edu/washington/escience/myria/operator/network/distribute/BroadcastDistributeFunction.java
```diff
@@ -0,0 +1,36 @@
...
+   */
+  @JsonCreator
+  public BroadcastDistributeFunction() {
+    super(new SinglePartitionFunction(), null);
+  }
+
+  /**
+   * @param numDestinations number of destination
+   */
+  public BroadcastDistributeFunction(final int numDestinations) {
```
jingjingwang: My understanding of Broadcast is that it duplicates a tuple to all the workers, so we don't need a list of them, while the mappings from partitions to destinations of a multiway join is manually specified in its `cellPartitition` so that Broadcast is not the correct one to use. However, if we want to extend `Broadcast` to broadcast to a subset of workers then yes we may need that, and I can envision how `HyperCubeShuffle` uses it.

src/edu/washington/escience/myria/operator/network/distribute/DistributeFunction.java
```diff
@@ -0,0 +1,82 @@
...
+import com.fasterxml.jackson.annotation.JsonSubTypes;
+import com.fasterxml.jackson.annotation.JsonSubTypes.Type;
+import com.fasterxml.jackson.annotation.JsonTypeInfo;
+import com.google.common.collect.Lists;
+
+import edu.washington.escience.myria.storage.TupleBatch;
+
+/**
+ * A dataset is distributed by two steps: First, using a partition function to generate a partition for each tuple;
+ * Second, mapping each partition to a set of destinations. A destination corresponds to an output channel, which is
```
jingjingwang: It is, I've changed the wording.

src/edu/washington/escience/myria/operator/network/distribute/DistributeFunction.java
```diff
@@ -0,0 +1,82 @@
...
+  private static final long serialVersionUID = 1L;
+
+  /** The partition function. */
+  protected PartitionFunction partitionFunction;
+
+  /** The mapping from partitions to destinations. */
+  protected List<List<Integer>> partitionToDestination;
+
+  /** number of destinations. */
+  protected int numDestinations;
```
jingjingwang: We can and they are the same, I just want to avoid recomputing the cardinality every time when a tuple batch is distributed by keeping the computed value there.

src/edu/washington/escience/myria/operator/network/distribute/IdentityDistributeFunction.java
```diff
@@ -0,0 +1,31 @@
+package edu.washington.escience.myria.operator.network.distribute;
+
+import com.fasterxml.jackson.annotation.JsonCreator;
+import com.fasterxml.jackson.annotation.JsonProperty;
+
+import edu.washington.escience.myria.util.MyriaArrayUtils;
+
+/**
+ * Implementation of a DistributeFunction that use the trivial identity hash. (i.e. a --> a) The attribute to hash on
+ * must be an INT column and should represent a workerID
```
jingjingwang: It is a distribute function so the attribute represents a worker ID.

src/edu/washington/escience/myria/operator/network/distribute/SinglePartitionFunction.java
```diff
@@ -0,0 +1,26 @@
+package edu.washington.escience.myria.operator.network.distribute;
+
+import javax.annotation.Nonnull;
+
+import com.fasterxml.jackson.annotation.JsonCreator;
+
+import edu.washington.escience.myria.storage.TupleBatch;
+
+/** return a fixed integer. */
+public final class SinglePartitionFunction extends PartitionFunction {
```
jingjingwang: I still prefer "Single" since there is only one partition.

src/edu/washington/escience/myria/operator/network/distribute/RoundRobinDistributeFunction.java
```diff
@@ -0,0 +1,28 @@
...
+   */
+  @JsonCreator
+  public RoundRobinDistributeFunction() {
+    super(new RoundRobinPartitionFunction(), null);
+  }
+
+  @Override
+  public void setNumDestinations(final int numWorker, final int numOperatorId) {
+    partitionFunction.setNumPartitions(numWorker);
+    partitionToDestination = MyriaArrayUtils.create2DVerticalIndexList(numWorker);
```
jingjingwang: That's right, also same for `Broadcast` if we want to broadcast to a subset of workers.

src/edu/washington/escience/myria/operator/network/distribute/BroadcastDistributeFunction.java
```diff
@@ -0,0 +1,36 @@
...
+    this();
+    partitionToDestination = MyriaArrayUtils.create2DHorizontalIndexList(numDestinations);
```
jingjingwang: That's right.

src/edu/washington/escience/myria/operator/network/RecoverProducer.java
```diff
@@ -25,20 +27,22 @@
...
+   */
   public RecoverProducer(
       final Operator child,
       final ExchangePairID operatorID,
-      final int collectConsumerWorkerID,
+      final int consumerWorkerID,
       final Producer oriProducer,
-      final int channelIndx) {
-    super(child, operatorID, collectConsumerWorkerID);
+      final int channelInx) {
```
senderista: `channelIdx`

src/edu/washington/escience/myria/operator/network/RecoverProducer.java
```diff
@@ -25,20 +27,22 @@
...
+    super(child, operatorID, consumerWorkerID);
     this.oriProducer = oriProducer;
```
senderista: `origProducer`

src/edu/washington/escience/myria/operator/network/distribute/HowDistributed.java
```diff
@@ -9,28 +9,32 @@
...
   public HowDistributed(@Nullable final DistributeFunction df, @Nullable final int[] workers) {
     this.df = df;
     if (workers != null) {
       this.workers = ImmutableList.copyOf(Ints.asList(workers));
     }
   }
 
-  /** Static function to create a HowPartitioned object.
+  /**
+   * Static function to create a HowPartitioned object.
```
senderista: `HowDistributed `

src/edu/washington/escience/myria/operator/network/distribute/HowDistributed.java
```diff
@@ -9,28 +9,32 @@
...
    *
    * @param df the distribute function.
    * @param workers the sequence of workers. *
-   * @return a new HowPartitioned reference to the specified relation. */
+   * @return a new HowPartitioned reference to the specified relation.
```
senderista: `HowDistributed`

src/edu/washington/escience/myria/operator/network/RecoverProducer.java
```diff
@@ -25,20 +27,22 @@
...
    * @param oriProducer the original producer.
-   * @param channelIndx the channel index that this operator is recovering for. * */
+   * @param channelInx the channel index that this operator is recovering for. *
+   */
   public RecoverProducer(
       final Operator child,
       final ExchangePairID operatorID,
-      final int collectConsumerWorkerID,
+      final int consumerWorkerID,
       final Producer oriProducer,
```
senderista: `origProducer`

src/edu/washington/escience/myria/operator/network/distribute/DistributeFunction.java
```diff
@@ -0,0 +1,82 @@
...
+      }
+    }
+    return result;
+  }
+
+  /**
+   * @param numWorker the number of workers to distribute on
+   * @param numOperatorId the number of involved operator IDs
+   */
+  public abstract void setNumDestinations(final int numWorker, final int numOperatorId);
```
jingjingwang: This is mostly for `HyperCubeShuffle`, but I agree that a separate setter may work better. I've changed the code.

src/edu/washington/escience/myria/operator/network/distribute/HashDistributeFunction.java
```diff
@@ -0,0 +1,40 @@
...
+   */
+  @JsonCreator
+  public HashDistributeFunction(@JsonProperty("indexes") final int[] indexes) {
+    super(new HashPartitionFunction(indexes), null);
+    this.indexes = indexes;
+  }
+
+  @Override
+  public void setNumDestinations(final int numWorker, final int numOperatorId) {
+    partitionFunction.setNumPartitions(numWorker);
```
jingjingwang: These numbers are not available when the constructor is called so we still need this method, but I've removed `partitionToDestination` from the constructor.

jingjingwang:
I added some more changes in https://github.com/uwescience/raco/tree/myria_broadcast but RACO needs more changes to fully deal with all these distribution functions.

src/edu/washington/escience/myria/operator/network/distribute/BroadcastDistributeFunction.java
senderista: This brings me to another point. Using the current 1:1 identification of partitions with workers it's trivial to infer the mapping of partitions to workers. But if partitions are now mapped to "destinations" which are actually ephemeral system-issued IDs whose meaning can't be reconstructed after the fact, that seems impossible. We can still inspect the list of worker IDs in `HowDistributed` to get the workers storing the relation (e.g., to schedule fragments for execution on those workers), but I assume we would also want to be able to reconstruct the mapping of partitions to workers by inspecting the `DistributeFunction` serialized in the catalog. But how do we do that when this mapping is intermediated by an ephemeral layer (partitions->channel IDs) that can't be reconstructed after the fact? And why persist this ephemeral mapping at all (as JSON serialized in the catalog) if it can't be meaningfully interpreted?

An alternative, I think, would be use the partition->channel ID mapping during execution but to generate an explicit mapping of partitions->workers for JSON serialization in the catalog.

src/edu/washington/escience/myria/operator/network/distribute/DistributeFunction.java
senderista: I think redundancy can only be justified by some cost savings, and I can't imagine that there's any measurable computational cost savings here. And if this is logically a derived quantity (even if you've cached the result), then it doesn't make sense for there to be a setter for it.

jingjingwang:
I ended up doing some refactoring in `QueryConstruct` for the three encodings that construct `DbQueryScan` instances and then noticed one thing: what is the purpose of having `storedRelationId` in `TableScanEncoding`? It's never been set in RACO. On the other hand, all the callers of https://github.com/uwescience/myria/blob/master/src/edu/washington/escience/myria/coordinator/MasterCatalog.java#L1145 pass null to this arg (except TableScanEncoding).

senderista:
I guess `storedRelationId` is supposed to distinguish between different relations with the same name (`RelationKey`). I think I've seen cases where MyriaX gets confused because an overwritten relation still has obsolete partitions stored in Postgres (on workers the newest version is no longer stored on), but I can't recall the details. Anyway, using `storedRelationId` *might* be helpful in such cases, although I'm not sure how just maintaining a single version of the relation in the catalog and always using the explicit set of workers from the catalog wouldn't solve this problem. Alternatively, maybe it was included to allow renaming a relation (I think pretty much all industrial DBs use this `name->objectID` approach to allow renaming objects). I don't know what the original design rationale was, though, since I wasn't around ;-)

senderista:
Anyway, I would support ripping out `storedRelationId` from `TableScanEncoding` and `getWorkersForRelation()`.

jingjingwang:
Another use case I can think of is: if one day we would like to add a mapping from logical names to physical representations then it might be useful, however I guess it would take a while for RACO to have it. Sounds great, I'll just remove it happily.

jingjingwang:
Hmm, I take back my word since distributing a relation in various ways is still useful, although not being utilized by RACO yet. I'll still reserved it then.

senderista:
Yeah, good point, we could still experiment with alternative physical representations in JSON query plans if they were supported in the scan encoding, even if not in RACO...

jingjingwang:
Done with the refactoring! Need reviews for 0738239. @senderista

senderista:
Feel free to merge as soon as the RACO PR (https://github.com/uwescience/raco/pull/548) is merged.

@BrandonHaynes if you're planning to run `myria-cluster update` on the demo cluster to pick up these changes, then we'll need to talk about schema upgrade. It might be simpler just to replace the cluster.

PR #864, opened 2016-12-30
Pipeline Parallel Ingest

jortiz16:
@parmitam recently requested the ability to pipeline parallel ingest with the rest of her query plans. This PR creates the encoding for the CSVFileScanFragmentOperator for Raco. Raco would then replace the load CSV encoding with something like the following:

```
encoding =
{
 "opType": "CSVFileScanFragment",
 "reader": dict({
                   "readerType": "CSV",
                   "schema": scheme_to_schema(self.scheme())
                 }, **self.options),
                 "source": {
                    "dataType": "S3",
                    "s3Uri": self.path
                 }
}
```

The tricky part about this operator is that we need to find the byte ranges for each worker once they are initialized at the Myria side. We can't do this through Raco.

There are 2 ways users could use parallel ingest (including features in this PR):

**1.** Through a REST call. Here, the byte ranges are computed and sent to the workers. By default, the server tries to use all available workers, but if the file is small it will only use a fraction of the workers. Users can override this by selecting a subset of workers themselves. 

**2.** Through Raco by submitting a query that loads a CSV file from S3. This will use the new CSVFileScanFragmentEncoding. Since we are using this encoding, the workers must be autonomous and figure out the byte ranges on their own. 

To separate between both options above, I had to introduce a ```flagAsRangeSelected``` to help determine whether the operator should determine the ranges or simply continue with the old logic if the ranges have been determined.  Although, I don't really like the idea of having to replicate the code that determines byte ranges on the CSVFileScanFragment operator. 

**To clarify:** it sounds like we can have each worker simply be autonomous and figure out the byte ranges on its own. If we do this, it doesn't seem like we can support the ability of having users selecting a subset of workers they wish to ingest to.


jortiz16:
Apologize for the messy diff in the init method. I simply added a ```if (!flagAsIncomplete)``` condition but the formatting rearranged my code and the diff thinks it's new.

senderista:
Sounds to me like we should include an optional public field `workerIds` of type `Set<Integer>` in the encoding and have it default to `args.getServer().getAliveWorkers()` like you do now in the `construct()` method. The user could optionally supply an arbitrary set of worker IDs in the JSON encoding, which would coincide with the `overrideWorkers` field on the fragment (I doubt we would expose this in MyriaL, though). Then each worker would independently figure out its own byte range from `AmazonS3Source.getFileSize()` and the set of all participating workers (corresponding to its sorted index in that set). We would then rip out the existing code in the REST API for doing this calculation and just always leave it to the workers. Sound reasonable?

jortiz16:
Thanks Tobin! I think this is all clear. 

I do have a question about the ```parallelIngestDataset```method  in the Server class. Since we will now move all the byte logic to the operator, how will we know which workers actually got data? 

I ask this because at the end of ```parallelIngestDataset``` we call the ```HowPartitioned``` method.  We can't rely on ```workersToIngest``` because in cases where the file to read is small, we might not use all the workers in the set.

senderista:
Good point! We can just continue to use the file size to calculate the required number of workers (and preferably randomly select them from the currently alive workers). The set of workers would be persisted in `CSVFileScanFragment` as I suggested, so each worker could calculate its own byte range (and its status as first/last worker) from its sorted rank in that set.  For RACO plans, this calculation would also need to be done in `CSVFileScanFragmentEncoding.construct()`, so I suppose it should be factored out into a helper in the `Server` class.

For RACO plans, we may need to introduce special logic in `QueryConstruct.setAndVerifyScans()` to ensure that `CSVFileScanFragment` operators (and their associated `DbInsert` operators) are only assigned to the set of workers encoded in the operator itself.

jortiz16:
There are a couple more issues:

**1)** @parmitam  needs the ability to explicitly determine the number of workers used for ingest.  She has a CSV file that contains S3 bucket URLs. When reading the file in parallel, each worker will then ingest the BLOB files specified by the URLs. The issue is that since her CSV file is small, our current implementation would only try to use 1 worker. 

I was thinking, maybe we can allow users to specify a fixed set of workers by including an optional boolean flag in the ```parallelIngestDataset ``` method? This option shouldn't be available on the Raco side. What do you think?

**2)** Parmita and I have also had issues adding ```AnonymousAWSCredentialsProvider``` to the  ```AWSCredentialsProviderChain``` in the ```AmazonS3Source``` class.  Assuming that users do not specify a role when launching myria-cluster, it makes sense to make it possible to read publicly available data from S3 by using anonymous credentials.

I've launched a myria cluster without a role and tried something like the following:

```
s3Client =
          new AmazonS3Client(
              new AWSCredentialsProviderChain(
                  new AnonymousAWSCredentialsProvider(), new DefaultAWSCredentialsProviderChain()));
```

This wouldn't work for me, but if I changed it to: 

```
s3Client = new AmazonS3Client(new AnonymousAWSCredentialsProvider())
```

This worked fine. Would you happen to know why this could be?

Thanks!

senderista:
1. If the CSV file is "tiny", why use parallel ingest to read it? If you want to ensure even distribution over workers, why can't you just use ordinary ingest followed by a shuffle with `RoundRobinPartitionFunction` or `WholeTupleHashPartitionFunction`?

2. `AnonymousAWSCredentialsProvider` is incompatible with `AWSCredentialsProviderChain`: https://issues.apache.org/jira/browse/HADOOP-13237?focusedCommentId=15314812&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-15314812. We need some way to determine whether we should initialize `AmazonS3Client` with `DefaultAWSCredentialsProviderChain` or `AnonymousAWSCredentialsProvider`. In the Hadoop S3A library I linked, they do this by checking the `fs.s3a.aws.credentials.provider` property. We could check the same property from Myria code, assuming this is an acceptable dependency, or we could introduce our own property. Given that we already have a hard dependency on YARN through REEF, I think depending on this property might be acceptable. The `myria-cluster create` command would have to set it to `DefaultAWSCredentialsProviderChain` or `AnonymousAWSCredentialsProvider` depending on whether or not the `--role` option was present (I would rather not deal with any AWS credentials other than from EC2 roles).

jortiz16:
1. Good point. So with @parmitam's implementation, she is processing the blob data as each row from the CSV is read. If she uses the RoundRobin implementation, it is slower since the coordinator would need to read the blob data for each row before sending it to the worker. Is that right Parmita?

2.  I see. Ok so in that case, in the `getS3Client()` method, I'll use the Hadoop configuration and call the `getPassword()` method on the property you described earlier. If no information is found, I'll initialize the client with the `DefaultAWSCredentialsProviderChain`. Does that sound about right?

senderista:
1. I may be misunderstanding how this processing works, but if each blob URL tuple from the CSV file is turned into a separate file download on the processing worker, then if that file download is done in some local operator, it should be performed on the worker where the blob URL has been shuffled, right? I.e, if the blob download operator is downstream of the shuffle, it will always execute on the worker which is the shuffle destination of the blob URL.

2. Yes, the AWS credential provider should be set in `core-site.xml`. (It isn't currently, but I'll change that.) See https://hortonworks.github.io/hdp-aws/s3-security/ for documentation of the `fs.s3a.aws.credentials.provider` property. Note that the fix for the Hadoop bug cited above won't go in until Hadoop 2.8, which isn't released yet, so anonymous access to S3 public data won't work from Hadoop, but it should work for our Myria code that directly uses `S3Client`. I'll change `myria-cluster create` to set `fs.s3a.aws.credentials.provider` to either `InstanceProfileCredentialsProvider` or `AnonymousAWSCredentialsProvider`, depending on whether the `--role` option is present.

jortiz16:
1. Yeah, I believe the issue is that the Blob download happens during the FileScan operator itself (which will run in the coordinator). If we decide to move the download to another operator, I'm not sure how this would work for pipelined query plans.. for example, how do we determine which operator should be doing the download? We can discuss this more during the meeting as well since @parmitam can fill in with more of the details.

2. Thanks for looking into this! I'll follow up once the change is made.

senderista:
1. Then I am completely confused about how this processing works. I don't understand how the FileScan operator can automatically download URLs from tuples in the file it is scanning, unless the operator itself has somehow been hacked to do that. Let's discuss in the meeting with @parmitam.

senderista:
Based on discussion with @mbalazin today, @parmitam will remove the blob-downloading magic from `FileScan` and we'll introduce a new Myria built-in expression provisionally called `download_blob()`, with input schema `[string]` and output schema `[blob]`. So a MyriaL query to download blobs from a CSV file of URLs could look like:
```
urls = load("https://uwdb.s3.amazonaws.com/sampleData/blobFiles.csv",
csv(schema(url:string)));
blobs = select download_blob(url) from urls;
store(blobs, blobs);
```
Note that because we now introduce an implicit round-robin shuffle after FileScan, the blob downloads are evenly distributed over workers.

jortiz16:
Thanks Tobin! Also, could you let me know when the `fs.s3a.aws.credentials.provider` is added so I could then modify `AmazonS3Source`? I'll include that feature to this PR.

senderista:
I backported the patch from https://issues.apache.org/jira/browse/HADOOP-13237 to our private Hadoop fork (which is used by gradle to build MyriaX). That means that the MyriaL `load()` statement will handle S3 URLs correctly if they point to public objects and the Myria node running the `DataInput` operator has no AWS credentials present. I updated `myria-cluster` to use only EC2 role credentials if they are present, and to otherwise use only anonymous credentials (the provider chain is never used). Myria AMIs have not yet been updated, so you'll need to use the `--unprovisioned` option to `myria-cluster create`.

senderista:
Oh, it might be helpful to see how the Hadoop code extracts an AWS credentials provider from Hadoop configuration: https://github.com/uwescience/hadoop/commit/6badcdb6b990014249121a42fe5d4091670bede1. I think in our case we don't need to worry about supporting `BasicAWSCredentialsProvider` (since I don't want to encourage users to store their AWS creds in Hadoop config). In fact, I think we can just fail if there's no valid class name in `fs.s3a.aws.credentials.provider`, rather than constructing a fallback provider chain. We could document that supported providers in `fs.s3a.aws.credentials.provider` are `InstanceProfileCredentialsProvider`, 
`EnvironmentVariableCredentialsProvider`, `AnonymousAWSCredentialsProvider`.

jortiz16:
Thanks Tobin, I followed your suggestion and looked at their implementation. It works! Currently, if no valid class name is found, my implementation silently fails. We could throw an IOException (which is what they do), but I wasn't sure if we wanted to propagate this exception up to other methods.

jortiz16:
According to the docs for getTrimmed() it seems like it does:
```
/**
   * Get the value of the <code>name</code> property as a trimmed <code>String</code>, 
   * <code>null</code> if no such property exists. 
   * If the key is deprecated, it returns the value of
   * the first key which replaces the deprecated key and is not null
   * 
   * Values are processed for <a href="#VariableExpansion">variable expansion</a> 
   * before being returned. 
   * 
   * @param name the property name.
   * @return the value of the <code>name</code> or its replacing property, 
   *         or null if no such property exists.
   */
  public String getTrimmed(String name) {
    String value = get(name);
    
    if (null == value) {
      return null;
    } else {
      return value.trim();
    }
  }
```

src/edu/washington/escience/myria/io/AmazonS3Source.java
```diff
@@ -95,7 +96,12 @@ public AmazonS3Client getS3Client() throws MyriaApiException {
         clientConfig.setMaxErrorRetry(3);
         s3Client = new AmazonS3Client(credentials, clientConfig);
       } catch (ClassNotFoundException e) {
-        throw new MyriaApiException(Status.INTERNAL_SERVER_ERROR, className + " not found ", e);
+        if (className == null) {
```
senderista: Probably should check for `null` and throw immediately after you assign `className`, since I'm not sure `Class.forName()` would actually throw a `ClassNotFoundException` for a `null` class name, and even if it did, I wouldn't want to assume that behavior would continue since it's not documented AFAICT.

PR #868, opened 2017-01-16
PerfEnforce on Myria

jortiz16:
This PR adds the PerfEnforce research component to Myria. 

I tried to keep PerfEnforce as self-contained as possible, but there are a couple features I did have to add to Myria in order to get it to work:
- Ability to create materialized views
- Ability to execute a SQL command directly on Postgres (uses the DbExecute operator)
- Ability to execute a SQL command and return the results in an array
- For TupleSink, I added the option to remove the column header

The entry point for the PerfEnforce code is through the ```PerfEnforceDriver``` class. This class focuses on two steps:

 1) Ingesting data, collecting statistics and generating the PSLA. This is all enclosed in the ```PerfEnforceDataPreparation``` class.
 2) The learning model that will schedule the incoming queries at different cluster sizes in order to meet the SLAs. This can be seen in the ```PerfEnforceOnlineLearning ``` class.

To actually see this work, the user will need to specify the "PerfEnforce" flag through myria-cluster, which will create a 12-node cluster. The web interface will be sending calls to the PerfEnforce API found in the ```PerfEnforceResource``` class.


src/edu/washington/escience/myria/api/encoding/PerfEnforceStatisticsEncoding.java
```diff
@@ -0,0 +1,40 @@
...
+ */
+public class PerfEnforceStatisticsEncoding {
+  @Required public String table_name;
+  @Required public long table_size;
+  @Required public String selectivity_predicate_001;
+  @Required public String selectivity_predicate_01;
+  @Required public String selectivity_predicate_1;
+
+  public PerfEnforceStatisticsEncoding(
+      final String table_name, final long table_size, final List<String> selectivityList) {
```
jingjingwang: I think we can just merge these two constructors since the second one is never called except by the first one.

src/edu/washington/escience/myria/perfenforce/PerfEnforceOnlineLearning.java
```diff
@@ -0,0 +1,396 @@
...
+  public void parsingOnlineFile(final int configurationIndex, final String predictionFileName)
+      throws PerfEnforceException {
+    try {
+      BufferedReader streamReader = new BufferedReader(new FileReader(predictionFileName));
+      String currentLine = "";
+      double nextQueryPrediction = 0;
+      while ((currentLine = streamReader.readLine()) != null) {
+        nextQueryPrediction = Double.parseDouble((currentLine.split(",")[0]).split(":")[1]);
+      }
+      streamReader.close();
```
jingjingwang: Put it in `finally` is safer. (Also there are several places with the similar issue.)

jingjingwang:
@jortiz16 I made a pass, although I have to admit that I didn't run any test :) But if it works on your side then I believe it's okay. BTW this branch has conflicts to be resolved.

jingjingwang:
Also just curious: is it going to run in Windows?

jingjingwang:
Another thing is: many Javadocs are missing. If you use Eclipse, I think it checks them automatically if the option is turned on.

src/edu/washington/escience/myria/perfenforce/PerfEnforceOnlineLearning.java
BrandonHaynes: Try-with-resources, not `finally` :)

jingjingwang: Apparently I need to update my knowledge on it :)

src/edu/washington/escience/myria/api/encoding/PerfEnforceTableEncoding.java
```diff
@@ -0,0 +1,55 @@
...
+/**
+ * This class returns an encoding describing a table's schema.
+ */
+public class PerfEnforceTableEncoding {
+  public RelationKey relationKey;
+  public String type;
+  public AmazonS3Source source;
+  public Schema schema;
+  public Character delimiter;
+  public Set<Integer> keys;
```
jortiz16: Not necessarily... should I follow a particular order?

src/edu/washington/escience/myria/api/encoding/PerfEnforceTableEncoding.java
```diff
@@ -0,0 +1,55 @@
...
+   * @param type can be either a "fact" or "dimension" type
+   * @param source the location of the table's data in S3
+   * @param schema the schema of the table
+   * @param delimiter the delimiter used on the table's data in S3
+   * @param keys the column id of the primary key
+   * @param corresponding_fact_key the column id of the corresponding foreign key, primarily only used for dimension tables
+   */
+  @JsonCreator
+  public PerfEnforceTableEncoding(
+      @JsonProperty(value = "relationKey", required = true) final RelationKey relationKey,
```
jortiz16: Not necessarily. I just didn't use the JSON encoding for ```PerfEnforceStatisticsEncoding```.  I will add it to keep it all consistent.

src/edu/washington/escience/myria/perfenforce/PerfEnforceUtils.java
```diff
@@ -0,0 +1,179 @@
...
+    return keyString;
+  }
+
+  /**
+   * Creates a new schema based on the given column ids
+   *
+   * @param keys the set of columns to concatenate
+   * @param schema the schema to create
+   */
+  public static Schema getAttributeKeySchema(final Set<Integer> keys, final Schema schema) {
```
jortiz16: I didn't know this existed! So much nicer now

src/edu/washington/escience/myria/perfenforce/PerfEnforceUtils.java
```diff
@@ -0,0 +1,179 @@
...
+   * @throws PerfEnforceException
+   */
+  public static String getMaxFeature(final Server server, String sqlQuery, final int configuration)
+      throws PerfEnforceException {
+
+    try {
+      String explainQuery = "EXPLAIN " + sqlQuery;
+      List<String> featuresList = new ArrayList<String>();
+      String maxFeature = "";
+      double maxCost = Integer.MIN_VALUE;
```
jortiz16: I'm pretty sure that the minimum cost for query execution from Postgres is 0.00. I can't find exact document to back this up, but I can't imagine there would be a query cost less than 0. In addition, I'm also only running SPJ queries. I'm not running indexes, stored procedures, or anything outside the ordinary :)

src/edu/washington/escience/myria/perfenforce/PerfEnforceUtils.java
```diff
@@ -0,0 +1,179 @@
...
+              explainQuery,
+              Schema.ofFields("explain", Type.STRING_TYPE),
+              PerfEnforceUtils.getWorkerRangeSet(configuration));
+
+      for (String currentWorkerFeatures : allWorkerFeatures) {
+        if (currentWorkerFeatures.contains("cost")) {
+          currentWorkerFeatures =
+              currentWorkerFeatures.substring(currentWorkerFeatures.indexOf("cost"));
+          currentWorkerFeatures = currentWorkerFeatures.replace("\"", " ");
+          String[] cmd = {
```
jortiz16: Thanks! This is better

src/edu/washington/escience/myria/perfenforce/PerfEnforceOnlineLearning.java
```diff
@@ -0,0 +1,396 @@
...
+              modifiedTrainingFileName,
+              predictionsFileName);
+
+      String[] arrayCommand =
+          new String[] {"java", "-classpath", MOAFileName, "moa.DoTask", moaCommand};
+
+      Process p = Runtime.getRuntime().exec(arrayCommand);
+
+      BufferedReader reader = new BufferedReader(new InputStreamReader(p.getInputStream()));
+      while ((reader.readLine()) != null) {}
```
jortiz16: Yes, I ran into an issue where the process would just hang even when I tried to flush the error stream. I ended up following suggestions from http://stackoverflow.com/questions/5483830/process-waitfor-never-returns.

jortiz16:
I had to do a rebase, but the only new commits are the last two ("addressing comments" and "modify based on rebase"). I still need to test to make sure that the new changes did not introduce bugs. For example, I had to change the PerfEnforceDataPreparation because I wasn't using the new ```BroadcastDistributeFunction```.

jortiz16:
Ok ready for another pass. I tested these changes on my own cluster deployment. Thanks!

jortiz16:
Hey @jingjingwang  would you happen to know why the coveralls test fails by any chance? I looked at some of the classes they reference, but it did not seem relevant to the changes I made...

src/edu/washington/escience/myria/api/PerfEnforceResource.java
```diff
@@ -0,0 +1,131 @@
...
+   * Begins generating a PSLA based on the input schema
+   *
+   * @param tableList describes all the tables and schema of the dataset
+   */
+  @POST
+  @Path("/preparePSLA")
+  public Response prepareData(final List<PerfEnforceTableEncoding> tableList)
+      throws PerfEnforceException, DbException, Exception {
+    server.getPerfEnforceDriver().preparePSLA(tableList);
+    return Response.noContent().build();
```
jingjingwang: I think this maps to HTTP return code 204, `Response.ok()` is more appropriate.

jingjingwang: I think this maps to HTTP return code 204. I'm not sure if 201 is more appropriate, if this POST creates any resource then 201 sounds better.

src/edu/washington/escience/myria/api/PerfEnforceResource.java
```diff
@@ -0,0 +1,131 @@
...
+  }
+
+  /**
+   * Sets the tier the user decides to purchase.
+   *
+   * @param tier the tier selected by the user
+   */
+  @POST
+  @Path("/setTier")
+  @Consumes(MediaType.MULTIPART_FORM_DATA)
```
jingjingwang: Using `MULTIPART_FORM_DATA` usually means that the input form contains some files, which seems to be not your case. (Same in other places)

src/edu/washington/escience/myria/api/PerfEnforceResource.java
```diff
@@ -0,0 +1,131 @@
...
+  /**
+   * Determines the SLA for a given query. The SLA returned is based on the tier purchased.
+   *
+   * @param querySQL the query from the user
+   */
+  @POST
+  @Path("/findSLA")
+  @Consumes(MediaType.MULTIPART_FORM_DATA)
+  public Response findSLA(@FormDataParam("querySQL") final String querySQL) throws Exception {
+    server.getPerfEnforceDriver().findSLA(querySQL);
```
jingjingwang: If it's only getting some data but not modifying any server state, then `GET` is more appropriate.

jortiz16: Maybe I'm misunderstanding, but  I have this as a `POST` since `/findSLA` requires data to be submitted.

jingjingwang: https://github.com/uwescience/myria/blob/master/src/edu/washington/escience/myria/api/QueryResource.java#L238 is an example of using GET with String parameters.

jortiz16: Thanks!

jingjingwang:
LGTM!

PR #870, opened 2017-01-21
Enable worker system GC

jingjingwang:
Add a REST API to Server to send system GC instructions to workers to trigger GCs. This can clean up memory without restarting the whole cluster.


senderista:
This looks fine but I'd like to discuss motivation before merging. Terminating the JVM on OOM exceptions from any thread is certainly the right thing to do in any case.

jingjingwang:
The motivation is simply to avoid restarting a new Myria cluster to run any query with a fresh start. Otherwise the query may get affected by previous queries.

senderista:
What's the intended usage pattern for the REST API? It's worker ID-specific, so do you intend to call it only on a particular "stuck" worker? What would be the effect if this message were sent to all workers simultaneously? Is that a case worth accommodating in the REST API?

senderista:
You were going to remove the REST API for this feature after our discussion, in favor of just issuing the `SYSTEM_GC` message automatically after each query is run, right?

jingjingwang:
Yep! I removed the old commit on REST API and added `System.gc()`.

senderista:
Oh, I thought the coordinator would send SYSTEM_GC messages to all workers on completion of a query. Why did you decide not to do that?

jingjingwang:
You're right, I don't know what I was thinking... I also need to introduce another message from workers to server to say that "my GC has completed".

jingjingwang:
I stopped in the middle when trying to add a message from workers to master saying "my GC has completed" because the implementation would look similar to `ADD_WORKER_ACK` or `REMOVE_WORKER_ACK`, but I felt that it's an overkill. However, I think the server is not supposed to run the next query until all worker GCs have completed, if we always trigger GCs after each query. So... now I'm just not sure what a good design is. What do you think?

senderista:
Is it possible to do this without coordination, i.e. the workers execute a GC before they return query status to the master?

jingjingwang:
Per discussion, I'll try making the workers send query end messages to master first then trigger GCs.

jingjingwang:
I realized that workers don't know if the current subquery is the last one, the only place to make such a decision is at master. So I made master send messages to workers for triggering GC when a query is done. I also added a REST API for triggering system GC for a specific worker, which is for my elastic memory project specifically.

src/edu/washington/escience/myria/parallel/Worker.java
```diff
@@ -159,13 +158,20 @@ public void run() {
...
                   if (LOGGER.isInfoEnabled()) {
                     LOGGER.info("received ADD_WORKER " + workerId);
                   }
                   connectionPool.putRemote(
                       workerId, SocketInfo.fromProtobuf(cm.getRemoteAddress()));
                   sendMessageToMaster(IPCUtils.addWorkerAckTM(workerId));
                   break;
+                case SYSTEM_GC:
+                  if (LOGGER.isInfoEnabled()) {
+                    LOGGER.info("Trigger System GC");
```
senderista: Given that we're not computing anything here, not sure why conditional is necessary...

jingjingwang: You're right, I forgot to downgrade it.

senderista:
LGTM, so I guess the REST API for worker GC is only called by your controller now?

jingjingwang:
Yes.

senderista:
temporarily reverted this merge after a regression: https://github.com/uwescience/myria/commit/33763d8279f74038c3ce83935da0c756638267fd

PR #873, opened 2017-01-27
refactor aggregate

jingjingwang:
- Unify aggregate operators: remove Single/MultiGroupByAggregate, make StreamingAggregate as a subclass of Aggregate, remove CountAllAggregator
- Introduce an abstraction of tuple hash table, make Aggregate/SymmetricHashJoin use it
- Make aggregators column-oriented: the internal state contains columns of "basic" aggregation ops (e.g. SUM, COUNT), and "complex" aggregation ops which emit results are mapped to them (e.g. AVG). Only internal aggregations are applied to input tuples. Results are generated by emit aggregations.


jingjingwang:
Work with https://github.com/uwescience/raco/pull/553.

PR #874, opened 2017-01-31
refactor aggregates

jingjingwang:
- Unify aggregate operators: remove Single/MultiGroupByAggregate, make StreamingAggregate as a subclass of Aggregate, remove CountAllAggregator
- Introduce an abstraction of tuple hash table, make Aggregate/SymmetricHashJoin use it
- Make aggregators column-oriented: the internal state contains columns of "basic" aggregation ops (e.g. SUM, COUNT), and "complex" aggregation ops which emit results are mapped to them (e.g. AVG). Only internal aggregations are applied to input tuples. Results are generated by emit aggregations.

Works with https://github.com/uwescience/raco/pull/553.


senderista:
Could you find a way to avoid re-opening a PR that's under review? I have to copy/paste a few dozen comments by hand into the new PR...

src/edu/washington/escience/myria/storage/Tuple.java
```diff
@@ -30,16 +27,26 @@
...
    * The data of the tuple.
    */
-  private final List<Field<?>> data;
+  private final MutableTupleBuffer data;
 
   /**
    * @param schema the schema of the tuple
    */
   public Tuple(final Schema schema) {
+    data = new MutableTupleBuffer(schema);
```
senderista: What is the rationale for replacing a simple `ArrayList<Field>` with `MutableTupleBuffer`? It appears that `MutableTupleBuffer` carries a lot of overhead for a single-row table, especially when we're creating lots of `Tuple`s (e.g. in your implementation of `Aggregator.addRow()`). Actually, I wonder if we need two versions of `Tuple`: an immutable one which is simply a `ReadableTable` view over selected columns in a selected row of a `ReadableTable`, and a mutable one which implements `AppendableTable` but holds only one row. It seems that combining these two responsibilities in a single class creates unnecessary overhead.

src/edu/washington/escience/myria/util/MyriaArrayUtils.java
```diff
@@ -153,4 +153,19 @@ private MyriaArrayUtils() {}
...
   }
+
+  /**
+   * Helper function that generates an array of the numbers in [start, start+length).
+   *
+   * @param start the size of the array.
+   * @param length the length of the array.
+   * @return an array of the numbers [start, start+length).
+   */
+  public static int[] range(final int start, final int length) {
```
senderista: We're on Java 8, so: `IntStream.range(start, start+length).toArray()`

test/edu/washington/escience/myria/operator/AggregateTest.java
```diff
@@ -144,10 +139,8 @@ private TupleBatch doAggOpsToCol(
 
   /**
    * Helper function to instantiate an aggregator and do the aggregation. Do not use if more than one TupleBatch are
-   * expected.
-   *
-   * This variant uses a SingleGroupByAggregate in order to do extra testing of the Aggregators by hitting a function
-   * that Aggregate does not use.
+   * expected. This variant uses a SingleGroupByAggregate in order to do extra testing of the Aggregators by hitting a
```
senderista: change reference to removed class `SingleGroupByAggregate`

src/edu/washington/escience/myria/api/encoding/OperatorEncoding.java
```diff
@@ -14,12 +14,11 @@
 
 /**
  * A JSON-able wrapper for the expected wire message for an operator. To add a new operator, two things need to be done.
- * 1. Create an Encoding class that extends OperatorEncoding.
- * 2. Add the operator to the list of (alphabetically sorted) JsonSubTypes below.
+ * 1. Create an Encoding class that extends OperatorEncoding. 2. Add the operator to the list of (alphabetically sorted)
```
jingjingwang: I wish there is a way to tell Eclipse not to reformat it... :)

jingjingwang:
For `TupleHashTable`, actually I'm expecting a lot of collisions from joins, since one join key may correspond to many tuples. Aggregates are fine. Since usually joins have much larger hash tables, I was not very motivated to do the optimization. I agree that it is a cool design if collisions are rare.

src/edu/washington/escience/myria/api/encoding/OperatorEncoding.java
```diff
@@ -50,7 +49,7 @@
   @Type(name = "LocalMultiwayProducer", value = LocalMultiwayProducerEncoding.class),
   @Type(name = "Merge", value = MergeEncoding.class),
   @Type(name = "MergeJoin", value = MergeJoinEncoding.class),
-  @Type(name = "MultiGroupByAggregate", value = MultiGroupByAggregateEncoding.class),
+  @Type(name = "Aggregate", value = AggregateEncoding.class),
```
jingjingwang: I wish there is a way to tell Eclipse to reformat it... :)

src/edu/washington/escience/myria/operator/Apply.java
```diff
@@ -128,7 +127,7 @@ protected TupleBatch fetchNextReady() throws DbException, InvocationTargetExcept
         // we elide 2 copies: one in `GenericEvaluator.evaluateColumn()` and one generating the Cartesian product.)
         if (onlySingleValuedExpressions()) {
           outputBuffer.absorb(
-              new TupleBatch(getSchema(), resultColumnsForTB, inputTuples.numTuples()));
+              new TupleBatch(getSchema(), resultColumnsForTB, inputTuples.numTuples()), true);
```
jingjingwang: This `true` is not for `new TupleBatch()`, it is for `outputBuffer.absorb()`.

src/edu/washington/escience/myria/operator/TupleHashTable.java
```diff
@@ -0,0 +1,131 @@
...
+
+/**
+ * An abstraction of a hash table of tuples.
+ */
+public final class TupleHashTable implements Serializable {
+  /** Required for Java serialization. */
+  private static final long serialVersionUID = 1L;
+
+  /** Map from hash codes to indices. */
+  private transient IntObjectHashMap<IntArrayList> keyToIndices;
```
jingjingwang: Exactly.

src/edu/washington/escience/myria/operator/SymmetricHashJoin.java
```diff
@@ -32,279 +27,93 @@
...
-   * The names of the output columns.
-   */
+  /** The names of the output columns. */
   private final ImmutableList<String> outputColumns;
-
-  /**
-   * The column indices for comparing of child 1.
-   */
+  /** The column indices for comparing of the left child. */
   private final int[] leftCompareIndx;
```
jingjingwang: After another though I think `leftCompareColumns` sounds better, it also matches `leftAnswerColumns`.

src/edu/washington/escience/myria/operator/agg/Aggregate.java
```diff
@@ -1,113 +1,155 @@
...
+  private void generateResult() throws DbException {
+    for (TupleBatch tb : groupStates.getData().getAll()) {
+      List<Column<?>> columns = new ArrayList<Column<?>>();
+      for (int i = 0; i < gfields.length; ++i) {
+        columns.add(tb.getDataColumns().get(i));
+      }
+      for (Aggregator agg : emitAggs) {
+        columns.addAll(agg.emitOutput(tb));
+      }
+      resultBuffer.absorb(new TupleBatch(getSchema(), columns), true);
```
jingjingwang: Same as above, this `true` is for `absorb()`.

src/edu/washington/escience/myria/operator/agg/BooleanAggregator.java
```diff
@@ -1,112 +1,78 @@
...
-   * @param value the value to be added.
-   * @param state the current state of the aggregate.
-   */
-  public void addBoolean(final boolean value, final Object state) {
-    BooleanAggState b = (BooleanAggState) state;
-    if (needsCount) {
-      b.count = LongMath.checkedAdd(b.count, 1);
+    ReplaceableColumn toCol = to.getColumn(stateCols[0], toRow);
+    final int inColumRow = to.getInColumnIndex(toRow);
+    switch (aggOp) {
```
jingjingwang: I'm not trying to add any new ops here, just refactoring.

src/edu/washington/escience/myria/operator/agg/AggregatorFactory.java
```diff
@@ -16,18 +15,44 @@
...
+
+  /**
+   * Generate internal aggregators, each aggregator corresponds to a column of the internal state.
+   *
+   * @param inputSchema the schema that incoming tuples will take.
+   * @param offset the starting column index of aggregators made by this factory in the state hash table.
+   * @return aggregators aggregators for internal state.
+   * @throws DbException if there is an error creating the aggregators
+   */
+  List<Aggregator> generateInternalAggs(final Schema inputSchema, final int offset)
```
jingjingwang: Do you think using expressions like in `Apply` would be a good idea? Actually `Aggregate` reminds me of `Apply` a lot especially after this refactoring.

senderista:
Right, that's why I suggested that maybe we should have different interfaces/impls for `TupleHashTable` based on whether keys are expected to reference unique indices. For aggregates this will be true, for joins no. I don't think we should necessarily assume that aggregate groups will be coarse-grained; someone will eventually run a query where that is not the case. Anyway, I opened https://github.com/uwescience/myria/issues/876 for this issue, so we can drop it from the PR :)

src/edu/washington/escience/myria/operator/agg/StringAggregator.java
```diff
@@ -1,141 +1,108 @@
...
-  /**
-   * @param aFieldName aggregate field name for use in output schema.
-   * @param aggOps the aggregate operation to simultaneously compute.
-   * @param column the column being aggregated over.
-   */
-  public StringAggregator(final String aFieldName, final AggregationOp[] aggOps, final int column) {
-    super(aFieldName, aggOps);
-    fromColumn = column;
+  protected StringAggregator(
+      final String inputName, final int column, final AggregationOp aggOp, final int[] stateCols) {
```
jingjingwang: Because an Aggregator only performs one op. The factory will generates multiple aggregators if we have multiple ops on one column.

src/edu/washington/escience/myria/operator/agg/UserDefinedAggregatorFactory.java
```diff
@@ -105,13 +106,12 @@ public UserDefinedAggregatorFactory(
...
 
     /* Set up the emitters. */
-    emitEvaluators = new ArrayList<>(emitters.size());
-    for (Expression expr : emitters) {
-      GenericEvaluator evaluator =
-          new GenericEvaluator(expr, new ExpressionOperatorParameter(null, stateSchema));
-      evaluator.compile();
-      emitEvaluators.add(evaluator);
-    }
+    // emitEvaluators = new ArrayList<>(emitters.size());
```
senderista: This can all be removed, right?

src/edu/washington/escience/myria/operator/agg/UserDefinedAggregatorFactory.java
```diff
@@ -123,8 +123,7 @@ public UserDefinedAggregatorFactory(
     }
     resultSchema = new Schema(types, names);
     return ImmutableList.of(
-        new UserDefinedAggregator(
-            initEvaluator, updateEvaluator, emitEvaluators, resultSchema, stateSchema, offset));
+        new UserDefinedAggregator(initEvaluator, updateEvaluator, resultSchema, stateSchema));
```
senderista: OK, so the `emitEvaluators` are now computed from `emitters` in the `Aggregate` operator?

src/edu/washington/escience/myria/operator/agg/UserDefinedAggregatorFactory.java
```diff
@@ -157,16 +156,35 @@ private ScriptEvalInterface getEvalScript(
...
+            .append(type.toJavaObjectType().getSimpleName())
+            .append("(")
+            .append(varCount)
+            .append("+")
+            .append(Expression.STATECOLOFFSET)
+            .append(", val")
+            .append(varCount)
+            .append(");\n");
+      } else {
+        // state.replaceType(I, stateRow, valI);
```
senderista: ?

src/edu/washington/escience/myria/operator/agg/UserDefinedAggregatorFactory.java
```diff
@@ -157,16 +156,35 @@ private ScriptEvalInterface getEvalScript(
...
-          .append(Expression.RESULT)
-          .append(".put")
-          .append(type.toJavaObjectType().getSimpleName())
-          .append("(")
-          .append(varCount)
-          .append(", val")
-          .append(varCount)
-          .append(");\n");
+      if (param.getStateSchema() == null) {
+        // state.putType(I, valI);
```
senderista: ?

senderista:
Unfortunately most of my comments on your last refactoring commit are hidden behind "Show outdated"...

src/edu/washington/escience/myria/operator/agg/StringAggregator.java
```diff
@@ -27,7 +27,7 @@ protected StringAggregator(
   private static final long serialVersionUID = 1L;
 
   /** Placeholder as MIN/MAX value of String. */
-  private static final String STRING_INIT_VALUE = "STRING_INIT_VALUE";
+  private static final String STRING_INIT_VALUE = "NULL";
```
jingjingwang: I have to remove the null check in `StringColumnBuilder` then.

src/edu/washington/escience/myria/expression/evaluate/GenericEvaluator.java
```diff
@@ -216,7 +217,7 @@ public EvaluatorResult evaluateColumn(final TupleBatch tb) throws InvocationTarg
         new TupleBuffer(Schema.ofFields(getExpression().getOutputName(), type));
     final WritableColumn resultsWriter = resultsBuffer.asWritableColumn(0);
     for (int rowIdx = 0; rowIdx < tb.numTuples(); ++rowIdx) {
-      eval(tb, rowIdx, countsWriter, resultsWriter, null);
+      eval(tb, rowIdx, tb, rowIdx, resultsWriter, countsWriter);
```
jingjingwang: This is somewhat a hack since this `tb` could be either the input tuple batch from child or a slice of the `TupleHashTable` state. However the stateful emit expressions only know to fetch data from the "state" even if the "state" serves more like an input tuple batch in this case. I added a comment and will think about how to make it more explicit.

src/edu/washington/escience/myria/operator/agg/StringAggregator.java
senderista: How about we use the Unicode BOM instead then:
```
private static final String STRING_INIT_VALUE = "\uFEFF";
```

src/edu/washington/escience/myria/operator/agg/Aggregate.java
```diff
@@ -138,20 +160,27 @@ protected void init(final ImmutableMap<String, Object> execEnvVars) throws DbExc
...
-      stateSchema = Schema.merge(stateSchema, factories[i].generateStateSchema(inputSchema));
+      internalAggs.addAll(factories[i].generateInternalAggs(inputSchema));
+      List<Expression> emits = factories[i].generateEmitExpressions(inputSchema);
+      Schema newStateSchema = factories[i].generateStateSchema(inputSchema);
+      stateSchema = Schema.merge(stateSchema, newStateSchema);
+      for (Expression exp : emits) {
+        GenericEvaluator evaluator =
+            new GenericEvaluator(
+                exp, new ExpressionOperatorParameter(newStateSchema, newStateSchema));
+        // if (evaluator.needsCompiling()) {
```
jingjingwang: I removed everything about `needsCompiling()` in the new commit. First, it doesn't hurt to always compile since it only happens in `init()`, and second, now we need to evaluate tuple-by-tuple when the input and the output tuple batches have different sizes, so we need the compiled expression. Found some old discussions in https://github.com/uwescience/myria/pull/548.

src/edu/washington/escience/myria/operator/agg/BooleanAggregator.java
```diff
@@ -17,19 +13,22 @@
...
+      final int fromRow,
+      final MutableTupleBuffer to,
+      final int toRow,
+      final int offset) {
     Objects.requireNonNull(from, "from");
-    ReplaceableColumn toCol = to.getColumn(stateCols[0], toRow);
+    ReplaceableColumn toCol = to.getColumn(offset, toRow);
     final int inColumRow = to.getInColumnIndex(toRow);
     switch (aggOp) {
       case COUNT:
```
jingjingwang: I don't plan to add them in this PR.

senderista: tracking this in https://github.com/uwescience/myria/issues/880

src/edu/washington/escience/myria/operator/agg/LongAggregator.java
```diff
@@ -63,42 +61,6 @@ public void addRow(
...
-          }
-          return ImmutableList.of(new DoubleColumn(ret, ret.length));
-        }
-      default:
-        throw new IllegalArgumentException(aggOp + " is invalid");
-    }
-  }
-
-  @Override
   protected boolean isSupported(final AggregationOp aggOp) {
```
jingjingwang: Here the "supported" means the exposed aggregate such as AVG but not internal aggregates from 'getInternalOps()'. `addRow()` is only called with an `aggOp` from 'getInternalOps()' . Do you think we should remove the check? It's more like a check for RACO. I feel that it doesn't hurt to double-check here.

src/edu/washington/escience/myria/operator/agg/PrimitiveAggregatorFactory.java
```diff
@@ -68,42 +74,68 @@ public PrimitiveAggregatorFactory(final Integer column, final AggregationOp aggO
...
     List<AggregationOp> cols = getInternalOps();
-    for (AggregationOp aggOp : aggOps) {
-      List<Integer> indices = new ArrayList<Integer>();
-      for (AggregationOp op : getInternalOps(aggOp)) {
-        indices.add(cols.indexOf(op) + offset);
+    List<Expression> exps = new ArrayList<Expression>();
+    for (int i = 0; i < aggOps.length; ++i) {
+      String name = aggOps[i].toString().toLowerCase() + "_" + inputSchema.getColumnName(column);
+      switch (aggOps[i]) {
+        case COUNT:
```
jingjingwang: I'm not sure if `getAggColumnType()` has what you want.

src/edu/washington/escience/myria/expression/evaluate/ExpressionEvalInterface.java
```diff
@@ -15,16 +15,18 @@
    * {@link edu.washington.escience.myria.operator.StatefulApply}. The variables will be fetched from the tuple buffer
    * using the rowId provided in {@link edu.washington.escience.myria.expression.VariableExpression}.
```
jingjingwang: It's an outdated comment, will remove in the new commit.

src/edu/washington/escience/myria/operator/agg/UserDefinedAggregatorFactory.java
jingjingwang: Yep, so Aggregators are not responsible for emitting results anymore. They only need to aggregate.

src/edu/washington/escience/myria/operator/agg/UserDefinedAggregatorFactory.java
jingjingwang: It's to show what expression is being constructed. I simply followed the format of the previous comment.

src/edu/washington/escience/myria/operator/agg/Aggregate.java
```diff
@@ -103,14 +110,29 @@ protected TupleBatch fetchNextReady() throws DbException {
...
-        columns.addAll(agg.emitOutput(tb));
+      int stateOffset = gfields.length;
+      int emitOffset = 0;
+      for (AggregatorFactory factory : factories) {
+        int stateSize = factory.generateStateSchema(inputSchema).numColumns();
+        int emitSize = factory.generateSchema(inputSchema).numColumns();
+        TupleBatch state =
+            tb.selectColumns(MyriaArrayUtils.range(stateOffset, stateOffset + stateSize));
+        for (GenericEvaluator eval : emitEvals.subList(emitOffset, emitOffset + emitSize)) {
+          columns.addAll(eval.evaluateColumn(state).getResultColumns());
```
jingjingwang: Yep I also have realized it. I don't know if we can assume that `Aggregate` would only have singe-valued emit expressions? For now it seems to be true, even when Python UDFs are involved. Will check with @parmitam .

jingjingwang:
I have addressed all comments and also adopted new changes in master. (Is there a way to make coverall pass? I didn't change much...) There are several things that I need to confirm with @parmitam, but the new commit is good to be reviewed first.

jingjingwang:
I will do #876 and #878 in other PRs.

jingjingwang:
I also cherry-picked two commits from #867 here since one is relevant.

test/edu/washington/escience/myria/operator/UserDefinedAggregatorTest.java
```diff
@@ -197,7 +197,7 @@ public void testRowOfMax() throws Exception {
         assertEquals(2, result.numColumns());
         assertEquals(Type.LONG_TYPE, result.getSchema().getColumnType(0));
         assertEquals(Type.STRING_TYPE, result.getSchema().getColumnType(1));
-        assertEquals(9999, result.getLong(0, 0));
+        assertEquals(10000, result.getLong(0, 0));
```
jingjingwang: I think the old test was wrong since the first column is the row index, and the expression is `+1`, so the last row should have row index to be 10000. I didn't check why the old test was wrong though.

src/edu/washington/escience/myria/expression/evaluate/GenericEvaluator.java
```diff
@@ -108,14 +129,36 @@ public void compile() throws DbException {
...
+   * @param inputRow index of the row that should be used for input data
+   * @param state additional state that affects the computation
+   * @param stateRow index of the row that should be used for state
    * @param result the table storing the result
    * @param count column storing number of results (null for single-valued expressions)
    * @throws DbException in case of error.
    */
   public void eval(
-      @Nonnull final ReadableTable input,
+      @Nullable final ReadableTable input,
```
jingjingwang: When an emit expression is evaluated, the `input` is null but the `state` is not, although the `state` works as an `input`.

src/edu/washington/escience/myria/expression/evaluate/GenericEvaluator.java
```diff
@@ -65,38 +78,46 @@ public GenericEvaluator(
...
-                  ExpressionEvalInterface.class,
-                  new String[] {
-                    Expression.INPUT,
-                    Expression.INPUTROW,
-                    Expression.STATE,
-                    Expression.STATEROW,
-                    Expression.RESULT,
-                    Expression.COUNT
-                  });
+      if (script.contains("append")) {
```
jingjingwang: I agree, I think we might be able to clean it when solving #878.

src/edu/washington/escience/myria/expression/evaluate/ExpressionEvalAppendInterface.java
```diff
@@ -0,0 +1,29 @@
...
+   *
+   * @param input the input tuple batch
+   * @param inputRow row index of the input tuple batch
+   * @param state optional state that is passed during evaluation
+   * @param stateRow row index of the state
+   * @param result a table storing evaluation results
+   * @param count a column storing the number of results returned from this row
+   */
+  void evaluate(
+      final ReadableTable input,
```
jingjingwang: It's actually `@Nullable` for emit expressions.

src/edu/washington/escience/myria/expression/evaluate/PythonUDFEvaluator.java
```diff
@@ -60,21 +52,20 @@
...
   private final PythonFunctionRegistrar pyFuncRegistrar;
-
   /** python worker process. */
   private PythonWorker pyWorker;
   /** index of state column. */
-  private final boolean[] isStateColumn;
-  /** tuple size to be sent to the python process, this is equal to the number of children of the expression. */
-  private int numColumns = -1;
-
+  private Set<Integer> stateColumns;
```
jingjingwang: It's simply for making `contains()` faster (in theory).

src/edu/washington/escience/myria/expression/evaluate/PythonUDFEvaluator.java
```diff
@@ -143,77 +128,70 @@ public void compile() {
...
-
-      // read result back
-      readFromStream(null, null, result, resultcol);
-
-    } catch (Exception e) {
-      throw new DbException(e);
+      ColumnBuilder<?> output = ColumnFactory.allocateColumn(outputType);
+      /* TODO: Leaving the count column to be null for now since since it's not used by Python evaluator for aggregate.
+       * A better design is to let the Aggregator emit two columns or even multiple columns. */
+      readFromStream(null, output);
```
jingjingwang: Then we need to write to the `WriteableColumn` in `PythonUDFEvaluator.eval()`.

src/edu/washington/escience/myria/operator/Apply.java
```diff
@@ -293,9 +293,7 @@ protected void init(final ImmutableMap<String, Object> execEnvVars) throws DbExc
       } else {
         evaluator = new GenericEvaluator(expr, parameters);
       }
-      if (evaluator.needsCompiling()) {
-        evaluator.compile();
-      }
+      evaluator.compile();
```
jingjingwang: Yep! It is finally reasonable.

src/edu/washington/escience/myria/expression/evaluate/PythonUDFEvaluator.java
```diff
@@ -60,21 +52,20 @@
...
+  /** column indices of child ops. */
   private int[] columnIdxs = null;
-
   /** Output Type of the expression. */
   private Type outputType = null;
-
   /** is expression a flatmap? */
   private Boolean isMultiValued = false;
+  /** Tuple buffers for each group key. */
+  private IntObjectHashMap<TupleBatchBuffer> buffer;
```
jingjingwang: It's a little bit awkward, since each entry in this `buffer` was designed to match a row of the `state` in `Aggregate`.  If we change it to a `TupleHashTable`, then we need to pass the values of keys of each row in `state` to this `TupleHashTable`, hash them, and check if the keys indeed match, in order to send all tuples of a group to Python. I think there is some CPU overhead. One the other hand, the memory overhead of this design comparing to using `TupleHashTable`, is the Java object overhead of a list of Columns per group key, which may not be very bad to me. I agree that it would be the best to get rid of this memory overhead without sacrificing time complexity.

src/edu/washington/escience/myria/expression/evaluate/ExpressionEvalAppendInterface.java
```diff
@@ -14,16 +17,16 @@
    *
    * @param input the input tuple batch
```
senderista: If this is optional (`@Nullable`) then it should be documented as such.

test/edu/washington/escience/myria/operator/AggregateTest.java
```diff
@@ -49,17 +49,9 @@
...
-   * All numeric aggs, in order: COUNT, MIN, MAX, SUM, AVG, STDEV
-   *
-   * MIN,MAX match the input type
-   *
-   * SUM is the big form (int->long) and (float->double) of the input type
-   *
-   * COUNT is always long
-   *
-   * AVG and STDEV are always double
+   * Ensure that the given Schema matches the expected numeric aggregate types for the given Type. All numeric aggs, in
```
senderista: This is messed up again.

jingjingwang:
I think coverall was messed up since there is no way to change the coverage by 0.8% through modifying a few lines of Java docs...

src/edu/washington/escience/myria/storage/TupleBuffer.java
```diff
@@ -349,4 +349,18 @@ public ReadableColumn asColumn(final int column) {
...
   }
+
+  /**
+   * Append the specified value to the specified destination column in this TupleBuffer from the source column.
+   *
+   * @param destColumn which column in this TB the value will be inserted.
+   * @param sourceColumn the column from which data will be retrieved.
+   * @param sourceRow the row in the source column from which data will be retrieved.
+   */
+  public final void put(
```
senderista: Long overdue!

jingjingwang: Yeah! I can't believe this interface was not there.

PR #884, opened 2017-03-03
added the query kill test

tomerk:
Requires the python api query kill to be added (https://github.com/uwescience/myria-python/pull/61)

@BrandonHaynes @senderista

### DIFF
--- a/integration_tests.py
+++ b/integration_tests.py
@@ -79,6 +79,23 @@ store(uploaddatatest, uploaddatatest);
             self.assertEqual(query.status, 'SUCCESS')
             self.assertListOfDictsEqual(query.to_dict(), expected)
 
+class KillQueryTest(MyriaTestBase):
+    def test(self):
+        # This is a long-running program
+        program = """
+x = [0 as exp, 1 as val];
+do
+  x = [from x emit exp+1 as exp, 2*val as val];
+while [from x emit max(exp) < 1000];
+store(x, powersOfTwo);
+"""
+        plan = self.connection.compile_program(program)
+        # We submit with submit_plan so as to not block
+        query = MyriaQuery.submit_plan(plan)
+
+        # We kill the query and make sure it was killed
+        query.kill()
+        self.assertEqual(query.status, 'KILLED')
 
 
 


BrandonHaynes:
uwescience/myria-python#61 has been merged and a new release of myria-python pushed to pypi.

integration_tests.py
```diff
@@ -79,6 +79,23 @@ def test(self):
...
+        program = """
+x = [0 as exp, 1 as val];
+do
+  x = [from x emit exp+1 as exp, 2*val as val];
+while [from x emit max(exp) < 1000];
+store(x, powersOfTwo);
+"""
+        plan = self.connection.compile_program(program)
+        # We submit with submit_plan so as to not block
+        query = MyriaQuery.submit_plan(plan)
```
senderista: It would be nice to just call a nonblocking `MyriaQuery` API with the program string. I guess this could be done in `myria-python` by either adding a new method to `MyriaQuery` that just generates the plan and calls `MyriaConnection.submit_query()`, or by adding a nonblocking version of `MyriaConnection.execute_program()`. @BrandonHaynes could you weigh in?

integration_tests.py
```diff
@@ -79,6 +79,23 @@ def test(self):
...
             self.assertListOfDictsEqual(query.to_dict(), expected)
 
+class KillQueryTest(MyriaTestBase):
+    def test(self):
+        # This is a long-running program
+        program = """
+x = [0 as exp, 1 as val];
+do
+  x = [from x emit exp+1 as exp, 2*val as val];
+while [from x emit max(exp) < 1000];
```
senderista: May as well make this limit humongous (say a billion) to avoid races, since we're going to kill it at the first opportunity.

integration_tests.py
BrandonHaynes: Yeah, we support asynchronous execution in `MyriaQuery` but not in the connection facade.  I'll update `MyriaConnection.execute_program` to support this and add an overload to `MyriaQuery.submit_query`.

BrandonHaynes: Nonblocking overload added in https://github.com/uwescience/myria-python/pull/62

Issue #876, opened 2017-02-10
Optimize `TupleHashTable` for aggregates


senderista:
Right now, `TupleHashTable` maps from an `int` hash value of an attribute group to an `int` array of matching row indexes. This is reasonable for joins, where any attribute group value is expected to have multiple matching tuples, but not for aggregates, where grouping tuples should be unique. In the latter case, we could optimize both memory usage and lookup efficiency by mapping hash codes directly to index values, handling collisions by using 64-bit hash codes (instead of the current 32-bit codes), and searching a small overflow table in the rare case of collisions (which for 64-bit codes wouldn't be expected until the number of grouping tuples approached `2^32`). Explicitly, the new table implementation would have type `LongIntHashMap`, where keys are 64-bit hash codes of attribute groups, values are row indexes of matching tuples, and collisions are indicated by the reserved value `-1` (since that cannot be a valid row index). The overflow table would have type `LongObjectHashMap<int[]>`.

Additionally, when indexes are expected to be unique, the existing  `getIndices()` interface which returns an `IntIterator` imposes unnecessary overhead. Instead, we should have a `getIndex()` method that just returns an `int` (with `-1` reserved to indicate absence). That could be implemented even without the data structure optimizations described above.

PR #889, opened 2017-04-19
Optimize `TupleHashTable` for aggregates

senderista:
fixes https://github.com/uwescience/myria/issues/876


src/edu/washington/escience/myria/operator/UniqueTupleHashTable.java
```diff
@@ -0,0 +1,167 @@
...
+      IntIterator iter = collidingIndexes.intIterator();
+      while (iter.hasNext()) {
+        int idx = iter.next();
+        if (TupleUtils.tupleEquals(tb, key, row, data, keyColumns, idx)) {
+          return idx;
+        }
+      }
+      assert false; // we should never get here
+    }
+    return index;
```
jingjingwang: What if the hash code exists (so no `ABSENT_VALUE`) but the key does not? Looks like we will return the index of another key with the same hash code.

src/edu/washington/escience/myria/operator/UniqueTupleHashTable.java
```diff
@@ -0,0 +1,167 @@
...
+      Preconditions.checkNotNull(collidingIndexes);
+      Preconditions.checkState(collidingIndexes.size() > 1);
+      IntIterator iter = collidingIndexes.intIterator();
+      while (iter.hasNext()) {
+        int idx = iter.next();
+        if (TupleUtils.tupleEquals(tb, key, row, data, keyColumns, idx)) {
+          return idx;
+        }
+      }
+      assert false; // we should never get here
```
jingjingwang: Similar to below, what if all these tuples in `collidingIndexes` do not have the same key as `key`

src/edu/washington/escience/myria/operator/UniqueTupleHashTable.java
```diff
@@ -0,0 +1,167 @@
...
+
+  /**
+   * Replace a matching tuple in the data table with the input tuple.
+   *
+   * @param tb the input tuple batch
+   * @param keyColumns the key columns
+   * @param row the row index of the input tuple
+   * @return if at least one tuple is replaced
+   */
+  public boolean replace(final TupleBatch tb, final int[] keyColumns, final int row) {
```
jingjingwang: It's never used, although it will be useful if we also make `SymmetricHashJoin` use `UniqueTupleHashTable`. (`TupleHashTable` has this method for `SymmetricHashJoin` to call.)

senderista: Right, I just duplicated it from `TupleHashTable` because there didn't seem to be any reason not to include it in the interface. (Arguably, they should both be explicitly implementing interfaces deriving from a common interface, which differ only on `getIndices()` vs. `getIndex()`.)

I prefer to keep the semantics of `UniqueTupleHashTable` clear in the interface (i.e., have only a `getIndex()` method), although certainly the overflow table idea would also work well when multiple tuples for a grouping key are rare but possible. I think we'd need to do a bit more refactoring in that case (probably factor out the `UniqueTupleHashTable` implementation and have a wrapper that implements the `getIndices()` interface and delegates to this implementation).

src/edu/washington/escience/myria/operator/UniqueTupleHashTable.java
senderista: You're absolutely right. I can't avoid this unless I check the grouping key with `tupleEquals()`. Good catch!

src/edu/washington/escience/myria/operator/UniqueTupleHashTable.java
senderista: You're right, this is very unlikely but possible! Will fix per below.

senderista:
As I mentioned in comments, I agree that it might be useful to reuse the _implementation_ for grouping keys with multiple result entries (although the current implementation does assume unique grouping keys and would need to be changed accordingly), but I think the interface should explicitly reflect the assumptions made about uniqueness of grouping keys (i.e., `getIndex()` vs. `getIndices()`). If we decide to reuse this implementation for cases in which uniqueness of grouping keys is not guaranteed (like joins in the absence of functional dependency information), we should change the interface accordingly.

senderista:
What do you think about a version of `DupElim` which uses 128-bit hash codes (so no collisions are expected until the number of tuples approaches `2^64`, which will never happen)? That should save a lot of space since no results ever need to be retained!

jingjingwang:
I think 64-bit is large enough since 2^32 is ~4G and 4G*(8+4)bytes = 48G already. Also even if it's 128-bit I guess we still need the logic for handling collisions. 

I like the idea of making it simple, `UniqueTupleHashTable` is good.

senderista:
Actually my point was that with 128-bit hash values, we *don't* need logic for handling collisions, if we can trust our hash function (and we can always use a cryptographic hash function if we don't trust a heuristic hash function like MurmurHash). Using the approximation `p = n^2/2m` (where `n` = number of items and `m` = number of bins), which works well when `p << 0.5`, we can conclude that the chance of a collision with say `n = 2^40` (a trillion items) is about `2^-72`, which is completely negligible. Any content-addressable storage system (e.g. git) makes the assumption that random collisions are statistically impossible, as do many cryptography schemes (like digital signatures).

64-bit hashes are inadequate for realistic data sizes since by the above formula, even for say 128M items (`2^27`) the probability of a collision is about `2^-11`, which is far from negligible.

BrandonHaynes:
Agree with @senderista -- if the crypto folks don't have to worry about collisions at this size, we don't either. Assuming that Murmur is approximately cryptographically secure, which I think it is.

senderista:
I don't think Murmur is cryptographically secure in the rigorous sense, but it has been empirically shown to have excellent distribution properties, which is probably enough for our purposes. If we're paranoid enough about collisions to use a real crypto hash function, MD5 should be sufficient, or SipHash if we care about performance.

jingjingwang:
My concern of using 128-bit is memory consumption. It's 64 bits more per key comparing to 64-bit, and let's say the chance of a collision is 1/10. The cost of each collision is the size of the Java array that stores indices, let's say it takes 50 bytes. Then 64-bit still uses less memory than 128-bit. (Note that we can't avoid storing tuples even for 128-bit because it's a hash table for keeping states.)

PR #890, opened 2017-05-01
Myria python worker

parmitam:
unflattening MyriaPythonWorker,  adding env variable to python process fix python_egg_cache.


senderista:
How are you getting REEF to deploy the `MyriaPythonWorker`  python module? Are you installing it by hand?

parmitam:
We'll have to do the pip install for myriapythonworker via myria-cluster, since reef doesn't support hierarchical libraries.

senderista:
I still don't think that having a separate package for the python worker code is a good idea, since it needs to be kept in sync with the rest of MyriaX. Flattening the namespace is not ideal, but why is it a dealbreaker? Also, having to manually install a package on the whole cluster isn't a great scenario for non-EC2 deployments.

parmitam:
Myria-python and raco - are tightly linked with myriaX  and are distributed via pip, separately, so this isn't breaking precedent. Flattening the namespace to require python worker be a single file -- makes the code uglier than it already is and harder to maintain -- having the myriaPythonWorker as a single pip installable library is a pretty clean solution. Lets figure out the reef installation issue separately.

senderista:
EC2 deployments aren't a problem if we keep this as a pip package: we can just install the pip package on each node from myria-cluster. My concern was more with non-EC2 deployments, as well as the possibility of the pip package getting out of sync with MyriaX. But your point about RACO and myria-python already being tightly coupled to MyriaX is well-taken.

senderista:
Per @BrandonHaynes suggestion, let's try simulating the nested namespace with nested classes, so we can keep this as self-contained module files.

PR #891, opened 2017-05-04
Fix state col offset

jingjingwang:
This PR fixes two problems: 1. when `STATE` is used as input for an aggregate with group by, the `StateExpression` needs to be aware of its column index offset. 2. add a special check for `count(*)` on an empty table to return 0, but do not return anything if there are more aggregates.

Travis will be fixed by https://github.com/uwescience/raco/pull/558.


src/edu/washington/escience/myria/expression/evaluate/PythonUDFEvaluator.java
```diff
@@ -115,6 +115,7 @@ public PythonUDFEvaluator(
     pyWorker = new PythonWorker();
     pyWorker.sendCodePickle(fs.getBinary(), columnIdxs.length, outputType, isMultiValued);
     buffer = new TupleBuffer(stateSchema);
+    groups = new IntObjectHashMap<IntArrayList>();
```
senderista: Was this not properly initialized before?

src/edu/washington/escience/myria/storage/TupleBuffer.java
```diff
@@ -361,6 +361,5 @@ public final void put(
       final int destColumn, final ReadableColumn sourceColumn, final int sourceRow) {
     checkPutIndex(destColumn);
     TupleUtils.copyValue(sourceColumn, sourceRow, this, destColumn);
-    columnPut(destColumn);
```
senderista: I don't immediately see the redundant call to `columnPut()`; where was it?

src/edu/washington/escience/myria/operator/agg/Aggregate.java
```diff
@@ -46,6 +47,8 @@
   protected final int[] gfields;
   /** Buffer for restoring results. */
   protected TupleBatchBuffer resultBuffer;
+  /** If we have outputted 0 as the count for count(*) on an empty relation. */
+  private boolean COUNTALL_ON_EMPTY;
```
senderista: This is not a constant, so shouldn't be named like one.

src/edu/washington/escience/myria/operator/agg/Aggregate.java
```diff
@@ -102,6 +105,16 @@ protected TupleBatch fetchNextReady() throws DbException {
       tb = child.nextReady();
     }
     if (child.eos()) {
+      if (!COUNTALL_ON_EMPTY
```
senderista: Would it be more broadly useful to have a boolean getter on `Operator` indicating whether an operator had ever returned any results (i.e. `nextReady() != null`)? If that could be reused elsewhere, it would be better than an ad-hoc flag.

src/edu/washington/escience/myria/operator/agg/Aggregate.java
```diff
@@ -102,6 +105,16 @@ protected TupleBatch fetchNextReady() throws DbException {
...
+          && gfields.length == 0
```
senderista: This check might be a little clearer if it were encapsulated in a helper.

src/edu/washington/escience/myria/expression/evaluate/PythonUDFEvaluator.java
jingjingwang: No, we missed it. (@parmita also found it)

src/edu/washington/escience/myria/storage/TupleBuffer.java
jingjingwang: It's in `TupleUtils.copyValue`, this function calls `putXXX()`s which call `columnPut()`

src/edu/washington/escience/myria/operator/agg/Aggregate.java
jingjingwang: Great idea, I switched to use `Operator.numOutputTuples` instead.

senderista:
This looks good, but to be super-nitpicky (and this is purely a matter of taste), I think the logic would be clearer if the helper only tested whether `CountAll` was the sole aggregate, so the check for emitting a `0` with no results would look like:
```
private boolean isCountAllOnlyAggregate() {
  return gfields.length == 0
      && internalAggs.size() == 1
      && internalAggs.get(0) instanceof PrimitiveAggregator
      && ((PrimitiveAggregator) (internalAggs.get(0))).aggOp == AggregationOp.COUNT;
}

if (getNumOutputTuples() == 0 && groupStates.numTuples() == 0 && isCountAllOnlyAggregate()) {
...
}
```

parmitam:
This has broken stateful agg for pythonUDFs. it should be sending a tuple list, but it is sending a tuple at a time again...

jingjingwang:
From the code https://github.com/uwescience/myria/blob/fix_state_col_offset/src/edu/washington/escience/myria/expression/evaluate/PythonUDFEvaluator.java#L181 it seems we still send a list of tuples. Is there a quick ref for the error log?

parmitam:
There is no failure, it just sends one tuple at a time. Also, it just send the tuple not the state.
( I am printing what is being sent to the python process so not sure what I can provide to help)

senderista:
Looks good from my end, I guess we'll have to wait for the Python issue to be resolved...

parmitam:
Python issue resolved :)
There were some changes in semantics that were throwing things off. everything works!
Looks great to me!

PR #892, opened 2017-05-08
`DupElim` implementation using only 128-bit hash codes

senderista:
Based on discussion in https://github.com/uwescience/myria/pull/889.

Summary: instead of storing unique tuples for deduplication of input tuples, we store their 128-bit hash codes. For a sufficiently high-quality hash function (currently `MurmurHash3`, but we could switch to a fast cryptographic hash function like `SipHash`), this yields vanishingly small probability of collisions (e.g. `2^-49` for a trillion input tuples). This assumption that collisions are statistically impossible for a sufficiently large hash space (about `N^3` for inputs of size `N`) is shared by many cryptographic systems as well as all content-addressable storage systems, such as `git`.

Since Java doesn't have a primitive type for 128-bit integers, we split the hash codes into two 64-bit integers and store them as the keys and values of a hash table. Until the number of input tuples approaches a billion or so, we don't expect collisions in the first 64-bit hash, so we separately handle collisions in an overflow table, with a special value in the main table to indicate collisions.

Note: it's currently unclear to me how to make this implementation fit into the `StreamingState` framework, like the current `DupElim` implementation. The hash codes don't seem to be a sensible output from the `StreamingState.exportState()` method, but I hope I'm misunderstanding something.

If possible, I'd like to benchmark this implementation against the current one for some very large datasets.


senderista:
FYI, there is a good discussion of hash collision probabilities (by an experienced crypto implementer) here: https://crypto.stackexchange.com/questions/39641/what-are-the-odds-of-collisions-for-a-hash-function-with-256-bit-output

jingjingwang:
How about using 64-bit hash code? My point is mostly for less memory consumption. It's 64 bits less per key comparing to 128-bit, and let's say the chance of a collision is 1/10. The cost of each collision is the size of the Java array that stores indices, let's say it takes 50 bytes. Then 64-bit still uses less memory than 128-bit.

You're right that `StreamingState` makes no sense for this `DupElimHashOnly`, maybe we can rename it to be `StatelessDupElim` or something like that.

senderista:
I don't quite understand your comments. The point of this implementation is to avoid storing tuples at all, just hash codes. In order to do that, we have to avoid collisions entirely, since there is no way to resolve collisions without comparing against the actual tuple values. We don't store indices because there is no table of tuples to index into. The design of mapping 64-bit hash codes (first half of the hash value) to 64-bit hash codes (second half of the hash value), and also to an overflow table (second half of the hash value) for collisions in the first hash code, is just a workaround for the fact that Java has no 128-bit integer primitive type. Otherwise we would simply store a single set of 128-bit integers. I can explain the design further in the Myria meeting today if you want.

jingjingwang:
Oh I see I missed a point, the cost of a collision includes storing the tuples.

jingjingwang:
I was thinking about #889 when making the above comments so I'm copying them there.

src/edu/washington/escience/myria/operator/DupElimHashOnly.java
```diff
@@ -0,0 +1,132 @@
...
+import com.gs.collections.impl.map.mutable.primitive.LongLongHashMap;
+import com.gs.collections.impl.map.mutable.primitive.LongObjectHashMap;
+
+import edu.washington.escience.myria.DbException;
+import edu.washington.escience.myria.Schema;
+import edu.washington.escience.myria.storage.TupleBatch;
+import edu.washington.escience.myria.util.HashUtils;
+
+/**
+ * A simple implementation of duplicate eliminate. It keeps the references to all the TupleBatches which contain unique
```
jingjingwang: Does it keep references to TBs? Also could you explain this operator with more details here

src/edu/washington/escience/myria/operator/DupElimHashOnly.java
```diff
@@ -0,0 +1,132 @@
...
+   * */
+  protected TupleBatch doDupElim(final TupleBatch tb) {
+    final int numTuples = tb.numTuples();
+    if (numTuples <= 0) {
+      return tb;
+    }
+    final BitSet toRemove = new BitSet(numTuples);
+    for (int i = 0; i < numTuples; ++i) {
+      final byte[] hashCodeBytes = HashUtils.hashRowBytes(tb, i);
+      Preconditions.checkArgument(
```
jingjingwang: I think the length would always be 16? Unless we change the hash function to be something else.

src/edu/washington/escience/myria/operator/DupElimHashOnly.java
```diff
@@ -0,0 +1,132 @@
...
+          hashCodeBytes.length == 16,
+          "Expected 16 bytes in hash code, found %d",
+          hashCodeBytes.length);
+      byte[] hashCodeBytes1 = Arrays.copyOfRange(hashCodeBytes, 0, 8);
+      byte[] hashCodeBytes2 = Arrays.copyOfRange(hashCodeBytes, 8, hashCodeBytes.length);
+      long hashCode1 = Longs.fromByteArray(hashCodeBytes1);
```
jingjingwang: nitpicking, `Longs.fromByteArray()` takes the first 8 bytes only so no need for `hashCodeBytes1`. Also I would like to even avoid creating `hashCodeBytes2` but I haven't found a good solution so far.

src/edu/washington/escience/myria/operator/DupElimHashOnly.java
```diff
@@ -0,0 +1,132 @@
...
+        Preconditions.checkState(collidingHashCodes.size() > 1);
+        LongIterator iter = collidingHashCodes.longIterator();
+        while (iter.hasNext()) {
+          long hc = iter.next();
+          if (hc == hashCode2) {
+            // duplicate found
+            toRemove.set(i);
+            break;
+          }
+        }
```
jingjingwang: Reaching here means that we have a new key that needs to be kept.

test/edu/washington/escience/myria/operator/DupElimTest.java
```diff
@@ -94,7 +94,8 @@
   @Test
   public void testDupElim() throws DbException {
     BatchTupleSource src = new BatchTupleSource(makeTestData());
-    StreamingStateWrapper dupElim = new StreamingStateWrapper(src, new DupElim());
+    // StreamingStateWrapper dupElim = new StreamingStateWrapper(src, new DupElim());
```
jingjingwang: Could you remove the comment? 
(I think it'd be the best if we have another test for the "Stateful" `DupElim` -- I can do it)

test/edu/washington/escience/myria/operator/DupElimTest.java
```diff
@@ -94,7 +94,8 @@
   @Test
   public void testDupElim() throws DbException {
```
jingjingwang: Maybe `testStatelessDupElim` assuming I'm gonna add another one

src/edu/washington/escience/myria/operator/DupElimHashOnly.java
```diff
@@ -0,0 +1,132 @@
...
+import edu.washington.escience.myria.DbException;
+import edu.washington.escience.myria.Schema;
+import edu.washington.escience.myria.storage.TupleBatch;
+import edu.washington.escience.myria.util.HashUtils;
+
+/**
+ * A simple implementation of duplicate eliminate. It keeps the references to all the TupleBatches which contain unique
+ * tuples.
+ * */
+public final class DupElimHashOnly extends UnaryOperator {
```
jingjingwang: Maybe 'StatelessDupElim'..?

src/edu/washington/escience/myria/operator/DupElimHashOnly.java
senderista: Done (got lazy with cut'n'paste).

src/edu/washington/escience/myria/operator/DupElimHashOnly.java
senderista: Well, it's not really stateless (the unique hash codes are the state)...

src/edu/washington/escience/myria/operator/DupElimHashOnly.java
senderista: I was verbose for clarity, but I made the code a bit more concise. (It would be a lot cleaner if Java had array slice syntax.)

src/edu/washington/escience/myria/operator/DupElimHashOnly.java
senderista: Good catch! I knew I would mess up somewhere...

src/edu/washington/escience/myria/operator/DupElimHashOnly.java
senderista: Yeah, the idea was that someone might change `HashUtils` to use another hash function that output fewer than 16 bytes (although then the second 8-byte range copy should fail). If this assert doesn't impact perf I'd like to leave it in, although I'll change the second range copy to hardcode 16 for the upper bound, since otherwise it could fail silently.

src/edu/washington/escience/myria/operator/DupElimHashOnly.java
senderista: Really, I'd like this to be the default implementation of `DupElim`, unless there's a reason it needs to preserve the `StreamingState` semantics. Do you see a reason not to do this?

senderista: I changed the name to `HashDupElim` for now (like "hash aggregator", "hash join", etc.).

test/edu/washington/escience/myria/operator/DupElimTest.java
senderista: Yeah, that change was just temporary to show this passes the unit tests, but I guess I'll just change it to `HashDupElim` since I assume that will be the new default.

test/edu/washington/escience/myria/operator/DupElimTest.java
senderista: I'll leave it as-is assuming this will be the default impl.

src/edu/washington/escience/myria/operator/DupElimHashOnly.java
senderista: I agree that creating a new `byte[]` just to copy the last 8 bytes sucks, but I haven't found a workaround either. It's annoying to have to produce garbage in what should be the fast path.

senderista: I found an ugly but effective workaround using `ByteBuffer` (ugly because `ByteBuffer`'s API is horrible).

PR #893, opened 2017-05-11
Mpw reconfig

parmitam:
moving MyriaPythonWorker to a single python file to redistribute with reef.


parmitam:
Tested and ready for review.

python/MyriaPythonWorker.py
```diff
@@ -0,0 +1,218 @@
...
+
+
+if __name__ == '__main__':
+    # Read a local port to connect to from stdin
+    java_port = int(sys.stdin.readline())
+    sock = socket.socket(socket.AF_INET, socket.SOCK_STREAM)
+    sock.connect(("127.0.0.1", java_port))
+    infile = os.fdopen(os.dup(sock.fileno()), "rb", 65536)
+    outfile = os.fdopen(os.dup(sock.fileno()), "wb", 65536)
+    exit_code = 0
```
senderista: Where is `exit_code` being returned from the process? Did you mean to call `sys.exit(exit_code)` somewhere?

python/MyriaPythonWorker.py
```diff
@@ -0,0 +1,218 @@
...
+
+    try:
+        main(infile, outfile)
+    except SystemExit as exc:
+        exit_code = exc.code
+    finally:
+        outfile.flush()
+        sock.close()
```
senderista: newline

python/MyriaPythonWorker.py
```diff
@@ -0,0 +1,218 @@
...
+        elif itemType == DataType.BLOB:
+            obj = self.loads(stream.read(length))
+        return obj
+
+    @classmethod
+    def read_tuple(self, stream, tuplesize):
+        datalist = []
+        for i in range(tuplesize):
+            # first element read type
+            elementType = self.read_int(stream)
```
senderista: be consistent with identifiers (`snake_case` not `camelCase` is preferred for locals in Python).

parmitam: Fixed.

python/MyriaPythonWorker.py
```diff
@@ -0,0 +1,218 @@
...
+            # Second read the length
+            length = self.read_int(stream)
+
+            if length == SpecialLengths.NULL or length == 0:
```
parmitam: it makes the code cleaner on the java side.

python/MyriaPythonWorker.py
```diff
@@ -0,0 +1,218 @@
...
+        elif output_type == DataType.BLOB:
+            self.write_int(DataType.BLOB, stream)
+            self.pickle_and_write(self, obj, stream)
+
+    @classmethod
+    def read_command(self, stream):
+        length = self.read_int(stream)
+
+        if length < 0:
+            print("command length is less than zero", file=sys.stderr)
```
parmitam: fixed.

python/MyriaPythonWorker.py
```diff
@@ -0,0 +1,218 @@
...
+            return None
+        s = stream.read(length)
+
+        if len(s) < length:
+            raise EOFError
+        unenc = base64.urlsafe_b64decode(s)
```
parmitam: because the registration happens over http, via myriaX REST api.

python/MyriaPythonWorker.py
parmitam: done

python/MyriaPythonWorker.py
parmitam: yes, fixed.

parmitam:
less than zero is typically an error, sent by Java side. We can't do anything about it in Python other than exit.

python/MyriaPythonWorker.py
```diff
@@ -127,7 +127,6 @@ def read_command(self, stream):
         length = self.read_int(stream)
 
         if length < 0:
-            print("Command length is less than zero!", file=sys.stderr)
             raise ValueError("Command cannot have a length zero.")
```
senderista: Again, the message is misleading since the exception is thrown when the length is _less than_ 0.

parmitam: fixed.

python/MyriaPythonWorker.py
```diff
@@ -0,0 +1,215 @@
...
+import base64
+
+
+class SpecialLengths(object):
+    PYTHON_EXCEPTION_THROWN = -3
+    END_OF_STREAM = -4
+    NULL = -5
+
+
+class DataType(object):
```
senderista: Why isn't this `enum` `0`-based (for compatibility with Java `enum`s)? Do you explicitly start the corresponding Java `enum` at `1`?

python/MyriaPythonWorker.py
```diff
@@ -0,0 +1,215 @@
...
+def main(in_file, out_file):
+    pickle_serializer = PickleSerializer()
+    try:
+        func = pickle_serializer.read_command(in_file)
+        tuple_size = pickle_serializer.read_int(in_file)
+        output_type = pickle_serializer.read_int(in_file)
+        is_flatmap = pickle_serializer.read_int(in_file)
+
+        if tuple_size < 1:
+            raise ValueError("size of tuple should not be less than 1 ")
```
senderista:
```
raise ValueError("Size of tuple should not be less than 1.")
```

python/MyriaPythonWorker.py
```diff
@@ -0,0 +1,215 @@
...
+
+        while True:
+            num_tuples = pickle_serializer.read_int(in_file)
+            if num_tuples == SpecialLengths.END_OF_STREAM:
+                exit(0)
```
senderista: If the effect of this `exit()` is exactly the same as a `break`, then you should use `break` (for the same reason you don't use `goto` if you can avoid it: it obscures control flow).

python/MyriaPythonWorker.py
```diff
@@ -0,0 +1,215 @@
...
+                        retval[i], out_file, output_type)
+            else:
+                pickle_serializer.write_with_length(retval, out_file, output_type)
+
+            out_file.flush()
+
+    except Exception:
+        try:
+            pickle_serializer.write_int(SpecialLengths.PYTHON_EXCEPTION_THROWN, out_file)
+            pickle_serializer.write_with_length(traceback.format_exc().encode(
```
senderista: Use the `DataType` value, not the magic number `5`, so I don't have to look it up to know what it means.

python/MyriaPythonWorker.py
```diff
@@ -0,0 +1,215 @@
...
+                "utf-8"), out_file, 5)
+            print(traceback.format_exc(), file=sys.stderr)
+        except IOError:
+            # JVM close the socket
```
senderista: "closed the socket"?

python/MyriaPythonWorker.py
```diff
@@ -0,0 +1,215 @@
...
+            pass
+        except Exception:
+            print("python process failed with exception: ", file=sys.stderr)
```
senderista: How about:
```
print("Python worker process failed with exception:\n{}".format(traceback.format_exc()), file=sys.stderr)
```

python/MyriaPythonWorker.py
```diff
@@ -0,0 +1,215 @@
...
+        protocol = 2
+        return cPickle.dumps(obj, protocol)
+
+    @staticmethod
+    def loads(obj):
+        return cPickle.loads(obj)
+
+    @staticmethod
+    def pickle_and_write(self, obj, stream):
+        serialized = self.dumps( obj)
```
parmitam: fixed.

python/MyriaPythonWorker.py
parmitam: done.

python/MyriaPythonWorker.py
parmitam: done.

python/MyriaPythonWorker.py
parmitam: done

python/MyriaPythonWorker.py
parmitam: done

python/MyriaPythonWorker.py
parmitam: done.

python/MyriaPythonWorker.py
parmitam: because the enum encodes lengths, anything >=0 is a valid length.

python/MyriaPythonWorker.py
```diff
@@ -0,0 +1,215 @@
...
+        stream.write(struct.pack("!d", value))
+
+    @staticmethod
+    def write_long(value, stream):
+        stream.write(struct.pack("!q", value))
+
+
+class PickleSerializer(Serializer):
+
+    @classmethod
```
parmitam: we don't follow that convention in Raco though, so I am going to stick to self.

python/MyriaPythonWorker.py
parmitam: done

python/MyriaPythonWorker.py
parmitam: any length >=0 is valid. Therefore, special lengths are all <0 and correspond to special lengths in MyriaX(Java).

python/MyriaPythonWorker.py
```diff
@@ -0,0 +1,215 @@
...
+        for i in range(tuplesize):
+            # first element read type
+            element_type = self.read_int(stream)
+            # Second read the length
+            length = self.read_int(stream)
+
+            if length == SpecialLengths.NULL or length == 0:
+                datalist.append(0)
+            # length is > 0, read the item now
+            elif length > 0:
```
parmitam: Yup. fixed.

parmitam:
All comments are addressed.

python/MyriaPythonWorker.py
senderista: I don't understand how the `DataType` enum encodes lengths...

parmitam: The corresponding java enum also starts at 1, otherwise none of this would work.

python/MyriaPythonWorker.py
BrandonHaynes: `cls` is the correct convention, RACO is wrong.   `@staticmethod` is best, if applicable.

BrandonHaynes: (Probably want to run `pylint` over this code, there are a number of other style errors that are trivial to fix.)

parmitam: I changed it to @staticmethod, but didn't change self to cls.
I'll run pylint.

python/MyriaPythonWorker.py
```diff
@@ -0,0 +1,216 @@
...
+        elif itemType == DataType.DOUBLE:
+            obj = self.read_double(stream)
+        elif itemType == DataType.BLOB:
+            obj = self.loads(stream.read(length))
+        return obj
+
+    @classmethod
+    def read_tuple(self, stream, tuplesize):
+        datalist = []
+        for i in range(tuplesize):
```
BrandonHaynes: `xrange`, if we know we're running in Python 2.

python/MyriaPythonWorker.py
```diff
@@ -0,0 +1,216 @@
...
+            print("Python worker process failed with exception:\n{}".format(traceback.format_exc()), file=sys.stderr)
+        exit(-1)
+
+
+if __name__ == '__main__':
+    # Read a local port to connect to from stdin
+    java_port = int(sys.stdin.readline())
+    sock = socket.socket(socket.AF_INET, socket.SOCK_STREAM)
+    sock.connect(("127.0.0.1", java_port))
+    infile = os.fdopen(os.dup(sock.fileno()), "rb", 65536)
```
BrandonHaynes: `infile` and `outfile` should be in a single `with` statement for deterministic closing.

python/MyriaPythonWorker.py
```diff
@@ -0,0 +1,216 @@
...
+
+    except Exception:
+        try:
+            pickle_serializer.write_int(SpecialLengths.PYTHON_EXCEPTION_THROWN, out_file)
+            pickle_serializer.write_with_length(traceback.format_exc().encode(
+                "utf-8"), out_file, DataType.BLOB)
+            print(traceback.format_exc(), file=sys.stderr)
+        except IOError:
+            # JVM closed the socket
+            pass
```
BrandonHaynes: Why is the exception being swallowed here?

parmitam: because the pythonworker might be terminated by JVM, the error would be displayed on the myriaX side.

python/MyriaPythonWorker.py
parmitam: we will want to move to python3 sooner than later so range should be fine. Also since the tuple size isn't large enough to be felled by inefficiency of range vs xrange.

python/MyriaPythonWorker.py
BrandonHaynes: A terminated JVM is not the only cause of an IO error.  Swallowing exceptions is almost universally bad design.  At least log it.

parmitam:
All the code review comments have been addressed.

python/MyriaPythonWorker.py
```diff
@@ -0,0 +1,218 @@
...
+            num_tuples = pickle_serializer.read_int(in_file)
+            if num_tuples == SpecialLengths.END_OF_STREAM:
+                break
+
+            tuple_list = []
+            for j in range(num_tuples):
+                tuple_list.append(pickle_serializer.read_tuple(in_file, tuple_size))
+
+            retval = func(tuple_list)
+            if is_flatmap > 0:
```
senderista: Please change to `if is_flatmap:`, since `is_flatmap` is implicitly boolean (can only take values in `{0,1}` per your format spec).

python/MyriaPythonWorker.py
senderista: I am already running `pylint` over `integration_tests.py` in Travis and plan to do the same on this code...

python/MyriaPythonWorker.py
```diff
@@ -0,0 +1,218 @@
...
+
+    except Exception:
+        try:
+            pickle_serializer.write_int(SpecialLengths.PYTHON_EXCEPTION_THROWN, out_file)
+            pickle_serializer.write_with_length(traceback.format_exc().encode(
+                "utf-8"), out_file, DataType.BLOB)
+            print(traceback.format_exc(), file=sys.stderr)
+        except IOError:
+            # JVM closed the socket
+            print("IOError: \n{}".
```
senderista: super-nitpick:
```
print("IOError:\n{}".
```

python/MyriaPythonWorker.py
```diff
@@ -0,0 +1,218 @@
...
+        elif item_type == DataType.DOUBLE:
+            obj = cls.read_double(stream)
+        elif item_type == DataType.BLOB:
+            obj = cls.loads(stream.read(length))
+        return obj
+
+    @classmethod
+    def read_tuple(cls, stream, tuplesize):
+        datalist = []
+        for i in range(tuplesize):
```
senderista: `for _ in range(tuplesize):`

python/MyriaPythonWorker.py
```diff
@@ -0,0 +1,218 @@
...
+        if tuple_size < 1:
+            raise ValueError("Size of tuple should not be less than 1.")
+
+        while True:
+            num_tuples = pickle_serializer.read_int(in_file)
+            if num_tuples == SpecialLengths.END_OF_STREAM:
+                break
+
+            tuple_list = []
+            for j in range(num_tuples):
```
senderista: `for _ in range(num_tuples):`

python/MyriaPythonWorker.py
```diff
@@ -0,0 +1,218 @@
...
+        elif output_type == DataType.BLOB:
+            cls.write_int(DataType.BLOB, stream)
+            cls.pickle_and_write(obj, stream)
+
+    @classmethod
+    def read_command(cls, stream):
+        length = cls.read_int(stream)
+
+        if length < 0:
+            raise ValueError("Command cannot be less than zero.")
```
senderista: `raise ValueError("Command length cannot be less than zero.")`

python/MyriaPythonWorker.py
```diff
@@ -0,0 +1,218 @@
...
+                "utf-8"), out_file, DataType.BLOB)
+            print(traceback.format_exc(), file=sys.stderr)
+        except IOError:
+            # JVM closed the socket
+            print("IOError: \n{}".
+                  format(traceback.format_exc()), file=sys.stderr)
+        except Exception:
+            print("Python worker process failed with exception:\n{}".
+                  format(traceback.format_exc()), file=sys.stderr)
+        exit(-1)
```
senderista: I think we need a `finally` clause at the level of the outer `try` block to ensure that both `out_file` and `sys.stderr` are flushed after an exception is caught and logged.

senderista: The builtin function `exit()` should not be used in programs (`sys.exit()` should be used instead): http://stackoverflow.com/questions/6501121/difference-between-exit-and-sys-exit-in-python

senderista: True. I guess `close()` (automatically called by the context manager on scope exit) will automatically call `flush()`, but I don't think even `close()` will be called if we explicitly call `sys.exit()`. Shouldn't we `raise` the exception after logging/persisting it? I think the interpreter will exit with nonzero status on an uncaught exception, so we shouldn't need to explicitly call `sys.exit()` unless we need to control the exact exit code. So I think the line `exit(-1)` (which should be calling `sys.exit()`) can just be replaced with `raise`, with no need for a `finally` clause given the `with` clause in the caller.

parmitam: I am already calling main() in a with clause, so we should not need a finally. I have a try because 
I want to log the errors in case of an exception. Changing exit to sys.exit

parmitam:
was this merged into master?

senderista:
This has been merged into `master`.

PR #897, opened 2017-06-07
Fixed the encoding for CSVFragmentTupleSource

jortiz16:
Fixed the encoding so we use the CSV parameters from the CsvTupleReader. Also, renamed `CSVFileScanFragment` to `CSVFragmentTupleSource` to match with `TupleSource`.


jortiz16:
Yeah, I recently opened a PR for RACO. It uses the new encoding name

PR #898, opened 2017-06-08
fixing parallel ingest single worker case

jortiz16:
Fixing an ingest bug for cases where parallel ingest only needed a single worker to read a file. With the old code, every worker would try to read the entire file (since `workerIndex` value was kept at 0). In other words, every worker would believe they were the first worker. 

Now, `workerIndex` starts at -1 and is only assigned to read the file if it is found in the workerIds array. If `workerIndex` is kept at -1, we mark `flagAsIncomplete` as True which then makes sure to not initialize the parser for the current worker.

### DIFF
--- a/src/edu/washington/escience/myria/operator/CSVFragmentTupleSource.java
+++ b/src/edu/washington/escience/myria/operator/CSVFragmentTupleSource.java
@@ -416,7 +416,7 @@ public class CSVFragmentTupleSource extends LeafOperator {
         }
       }
       if (workerIndex >= 0) {
-        boolean isLastWorker = workerIndex == workerIds.length - 1;
+        boolean isLastWorker = (workerIndex == workerIds.length - 1);
         long startByteRange = currentPartitionSize * workerIndex;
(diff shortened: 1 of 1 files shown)


src/edu/washington/escience/myria/operator/CSVFragmentTupleSource.java
```diff
@@ -409,24 +409,28 @@ protected void init(final ImmutableMap<String, Object> execEnvVars) throws DbExc
...
       for (int i = 0; i < workerIds.length; i++) {
         if (workerID == workerIds[i]) {
           workerIndex = i;
         }
       }
-      boolean isLastWorker = workerIndex == workerIds.length - 1;
-      long startByteRange = currentPartitionSize * workerIndex;
-      long endByteRange;
+      if (workerIndex >= 0) {
+        boolean isLastWorker = workerIndex == workerIds.length - 1;
```
senderista: Parens around the condition would be more readable (I realize this was in the original).

PR #899, opened 2017-06-10
Include REEF container exceptions in Myria query response

senderista:
This change ensures that all REEF Evaluator/Context/Task failure exceptions will be properly rethrown by the coordinator and show up in the query response. Unfortunately, OOM failures that result from the container being forcibly terminated by the YARN NodeManager rather than a Java OOM exception are not attributed clearly, but this is still an improvement.

Example of an evaluator that was forcibly terminated for exceeding its memory quota:
```
edu.washington.escience.myria.DbException: Query #9.0 failed: org.apache.reef.exception.EvaluatorException: Evaluator [container_1497062636339_0001_01_000004] is assumed to be in state [RUNNING]. But the resource manager reports it to be in state [FAILED]. This means that the Evaluator failed but wasn't able to send an error message back to the driver. Task [1] was running when the Evaluator crashed.
	at edu.washington.escience.myria.parallel.MasterSubQuery$WorkerExecutionInfo$2.operationComplete(MasterSubQuery.java:119)
```
Example of an evaluator that triggered a Java OOM exception:
```
Suppressed: edu.washington.escience.myria.DbException: Worker #2 failed: java.lang.OutOfMemoryError: Java heap space
		... 20 more
	Caused by: edu.washington.escience.myria.parallel.WorkerFailedException: java.lang.OutOfMemoryError: Java heap space
		... 20 more
	Caused by: java.lang.OutOfMemoryError: Java heap space
		at java.lang.String.toCharArray(String.java:2899)
```


src/edu/washington/escience/myria/parallel/QueryManager.java
```diff
@@ -595,12 +595,13 @@ protected void killAll() {
    * queries may be killed or be told that the worker has died.
    *
    * @param workerId the worker that died.
+   * @param cause the exception describing why the worker died.
    */
-  protected synchronized void workerDied(final int workerId) {
+  protected synchronized void workerDied(final int workerId, final Throwable cause) {
```
jingjingwang: Add @Nullable as in other places?

src/edu/washington/escience/myria/parallel/WorkerFailedException.java
```diff
@@ -0,0 +1,31 @@
...
+
+  /**
+   * Default constructor.
+   * */
+  public WorkerFailedException() {}
+
+  /**
+   * @param msg the message describing this Exception.
+   * */
+  public WorkerFailedException(final String msg) {
```
jingjingwang: Looks like it's never used.

PR #900, opened 2017-07-12
Add blob literal support

### LINKED ISSUE #894: Properly support BLOB_TYPE literals
When https://github.com/uwescience/raco/issues/555 is resolved, we need to properly handle byte literals in MyriaX. In particular, `/myria/src/edu/washington/escience/myria/expression/ConstantExpression.java` should be changed to decode a `ConstantExpression` with type `BLOB_TYPE` to a `byte[]`, not a `ByteBuffer`, and a blob literal in MyriaL should be serialized in RACO to a base64 string (which is the only portable way to serialize byte sequences to JSON).

senderista:
Fixes #894. This depends on blob literal support in RACO, which has already been merged (https://github.com/uwescience/raco/pull/562).


jingjingwang:
I only have a high-level question: why would we need `BitSetExpression`? If this change is only for testing purpose then I don't see why we need to transfer a byte array to a `BitSet` and eventually to  a `boolean[]`.

senderista:
`BitsetExpression` was just a "cool idea" and seemed like a nice way to test blob types with a flatmap apply. I don't have a specific use case for it, although I hope there will eventually be one :)

jingjingwang:
Okay, how about calling it something like `BlobToBooleanExpression` instead of `BitSet`? A 'BitSet' is only used internally in the Java expression. (I wish there is a way to transfer a `ByteBuffer` to a `boolean[]` without using `BitSet`)

senderista:
Actually the name was deliberate; the other choices might have been "bitvector" or "bit array". The idea is that the input is interpreted as a bitvector (sequence of booleans) rather than a sequence of bytes. I think a name like BlobToBoolean wouldn't convey that interpretation clearly.

PR #902, opened 2017-07-14
Python io

parmitam:
marshall python exceptions correctly.

### DIFF
--- a/python/MyriaPythonWorker.py
+++ b/python/MyriaPythonWorker.py
@@ -125,8 +125,9 @@ class PickleSerializer(Serializer):
             cls.write_int(DataType.BLOB, stream)
             cls.pickle_and_write(obj, stream)
         elif output_type == DataType.EXCEPTION:
+            assert type(obj) is str
             cls.write_int(len(obj), stream)
-            stream.write(str(obj))
+            stream.write(obj)
 
     @classmethod
     def read_command(cls, stream):
--- a/python/MyriaPythonWorker.py
+++ b/python/MyriaPythonWorker.py
@@ -20,6 +20,7 @@ class DataType(object):
     FLOAT = 3
     DOUBLE = 4
     BLOB = 5
+    EXCEPTION = 6
 
 
 class Serializer(object):
@@ -123,6 +124,9 @@ class PickleSerializer(Serializer):
         elif output_type == DataType.BLOB:
             cls.write_int(DataType.BLOB, stream)
             cls.pickle_and_write(obj, stream)
+        elif output_type ==DataType.EXCEPTION:
+            cls.write_int(len(obj), stream)
+            stream.write(str(obj))
 
     @classmethod
     def read_command(cls, stream):
@@ -198,7 +202,7 @@ def main(in_file, out_file):
             pickle_serializer.write_int(
                 SpecialLengths.PYTHON_EXCEPTION_THROWN, out_file)
             pickle_serializer.write_with_length(traceback.format_exc().encode(
-                "utf-8"), out_file, DataType.INT)
+                "utf-8"), out_file, DataType.EXCEPTION)
             print(traceback.format_exc(), file=sys.stderr)
         except IOError:
             # JVM closed the socket
--- a/python/MyriaPythonWorker.py
+++ b/python/MyriaPythonWorker.py
@@ -124,7 +124,7 @@ class PickleSerializer(Serializer):
         elif output_type == DataType.BLOB:
             cls.write_int(DataType.BLOB, stream)
             cls.pickle_and_write(obj, stream)
-        elif output_type ==DataType.EXCEPTION:
+        elif output_type == DataType.EXCEPTION:
             cls.write_int(len(obj), stream)
             stream.write(str(obj))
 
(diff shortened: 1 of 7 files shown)


parmitam:
exception was being incorrectly marshaled from python.
Also cleaned up some extra exceptions from contant and python evaluators

python/MyriaPythonWorker.py
```diff
@@ -123,6 +124,9 @@ def write_with_length(cls, obj, stream, output_type):
         elif output_type == DataType.BLOB:
             cls.write_int(DataType.BLOB, stream)
             cls.pickle_and_write(obj, stream)
+        elif output_type ==DataType.EXCEPTION:
```
senderista: space after `==`

python/MyriaPythonWorker.py
```diff
@@ -123,6 +124,9 @@ def write_with_length(cls, obj, stream, output_type):
...
+            cls.write_int(len(obj), stream)
```
senderista: I don't think this is quite correct in general. The only caller converts a `unicode` object (return value of `format_exc()`) to a UTF-8 `str`, and then passes it to this method, so `len()` works fine. But in general, if this method accepts an arbitrary Python object (i.e., anything that `str()` can format), there's no reason to think that `len(str(obj)) == len(obj)`. So if you want to be able to marshal any object with `DataType.EXCEPTION`, and as long we're stuck with Python 2, I think it would be more consistent for this method to call `unicode(obj).encode("utf-8")` and write the resulting bytes to `stream`. Then the sole caller should not bother encoding the result of `format_exc()`, so should just call `pickle_serializer.write_with_length(traceback.format_exc(), out_file, DataType.EXCEPTION)`.

python/MyriaPythonWorker.py
```diff
@@ -198,7 +202,7 @@ def main(in_file, out_file):
             pickle_serializer.write_int(
                 SpecialLengths.PYTHON_EXCEPTION_THROWN, out_file)
             pickle_serializer.write_with_length(traceback.format_exc().encode(
-                "utf-8"), out_file, DataType.BLOB)
+                "utf-8"), out_file, DataType.EXCEPTION)
```
senderista: See above.

src/edu/washington/escience/myria/expression/evaluate/PythonUDFEvaluator.java
```diff
@@ -224,12 +226,13 @@ public void readFromStream(final WritableColumn count, final WritableColumn resu
       for (int i = 0; i < c; i++) {
         // then read the type of tuple
         int type = dIn.readInt();
-        // if the 'type' is exception, throw exception
+        // Signals that an exception has been thrown in python
         if (type == MyriaConstants.PythonSpecialLengths.PYTHON_EXCEPTION.getVal()) {
           int excepLength = dIn.readInt();
```
senderista: A little more consistency would be nice, say `excLen`, `exc`.

src/edu/washington/escience/myria/operator/agg/StreamingAggregate.java
```diff
@@ -41,7 +40,7 @@ public StreamingAggregate(
    *
    * @throws DbException if any error occurs.
    * @return result tuple batch
-   * @throws IOException
+   *
```
senderista: delete blank line

src/edu/washington/escience/myria/expression/evaluate/ConstantEvaluator.java
```diff
@@ -1,6 +1,8 @@
 package edu.washington.escience.myria.expression.evaluate;
 
+import java.io.IOException;
```
senderista: unused import?

python/MyriaPythonWorker.py
parmitam: This is exceptions, not any python objects, typically a  string.  we convert this to a string with UTF_8 standard character set on the java side. 
I believe this should be fine.

src/edu/washington/escience/myria/expression/evaluate/ConstantEvaluator.java
parmitam: I think I fixed this,  multiple commits :(

src/edu/washington/escience/myria/operator/agg/StreamingAggregate.java
parmitam: done.

src/edu/washington/escience/myria/expression/evaluate/PythonUDFEvaluator.java
parmitam: done.

python/MyriaPythonWorker.py
parmitam: done.

python/MyriaPythonWorker.py
senderista: If the object is always a `str`, then there is no reason to convert it with `str()`. You should just write the bytes directly. If you need to ensure that it's always a `str`, then add an assert `type(obj) is str`. If the object could potentially be another type, then we should always explicitly call `unicode(obj).encode("utf-8")` in the method.

senderista: BTW, if the object is always expected to be the output of `traceback.format_exc()`, then it is a `unicode` object, so we should assert `type(obj) is unicode` and just call `obj.encode("utf-8")`.

PR #905, opened 2017-07-20
Partitioning order integration test

senderista:
This test depends on the RACO fix that I just merged (fails without it): https://github.com/uwescience/raco/pull/565


jortiz16:
This looks good to me. Great catch on that RACO bug!

PR #910, opened 2017-08-30
specify a size to a column builder

jingjingwang:
By default, a column builder reserves 10k data except `BlobColumnBuilder`. However, when a list of columns (including a blob column) is allocated, the blob column is of size 1 while other columns are of size 10k. This caused OOM when we modified `ApplyTest.testApply()` by adding a blob column to its schema -- since each `TupleBatch` is now of size 1, the test created 20k `TupleBatch`es, each has a few columns of size 10k (and one blob column of size 1). 
Fix this bug by passing the size to a column builder, instead of getting its size based on its type internally.

### DIFF
--- a/src/edu/washington/escience/myria/MyriaConstants.java
+++ b/src/edu/washington/escience/myria/MyriaConstants.java
@@ -12,24 +12,16 @@ public final class MyriaConstants {
    */
   public static final String SYSTEM_NAME = "Myria";
 
-  /**
-   * 1 kb.
-   */
+  /** 1 KB. */
   public static final int KB = 1024;
 
-  /**
-   * 1 mb.
-   */
+  /** 1 MB. */
   public static final int MB = 1024 * KB;
 
-  /**
-   * 1 gb.
-   */
+  /** 1 GB. */
   public static final int GB = 1024 * MB;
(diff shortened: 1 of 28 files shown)


jingjingwang:
For 1), yes the purpose of introducing `blob` was to handle large binary objects such as images in the first place. We expected that small objects can be encoded using other types (`String`, for example). That's why we hard-coded the batch size to be 1 when `blob` is involved. It's not a complete design for sure, which goes to 2): I agree that we would need to add interfaces to `Operator` to specify the batch sizes of its output tuple batches, but then each operator needs to know how to handle cases when input and output batch sizes are different. There will be some buffering/copying functions to add. At that time, we didn't have a use case to motivate this big change so we didn't implement it. 

For 3), it's mainly because that `byte[]` doesn't implement `Comparable`. Certainly we can also implement our own wrapper for `byte[]` with `Comparable`, but it seems unnecessary if `ByteBuffer` already has it. It may also have efficient I/O (we didn't benchmark anything tho).

parmitam:
looks good!

senderista:
OK, I guess we can address the design warts when we have a compelling use case.

